PR #4, opened 2025-02-01
Refactor Distance Functions for Improved Error Handling and Performance

LVivona:
#### Summary:
This PR refactors the distance functions (`manhattan_distance`, `hamming_distance`, `jaccard_distance`) to improve error handling and enhance performance. The changes focus on:
- **Better Error Handling**: Using `thiserror::Error` for more descriptive and actionable error messages.
- **Performance Improvement**: Replacing iterator methods (`iter()`, `map()`, etc.) with traditional `for` loops, as they are generally faster in performance-critical scenarios. This refactor prioritizes simplicity and efficiency.

#### Details of Changes:

1. Error Handling Improvements:
- Previously, error handling was done by returning static strings.
- With this update, errors are now returned using thiserror, which provides more structured error types and automatically implements Debug, Display, and Error traits.
- Errors are returned using the DistanceError enum, offering better context, traceability, and clearer messaging.
- This change allows for more robust error handling, with specific error variants for empty inputs and mismatched array lengths.

**NOTE** I noticed that in another part of your code, an error struct is used, but only one type of error is possible in that case. In such situations, instead of using Result, it may be more appropriate to use an ``Option``, as there is only one possible error to handle.

2. **Performance Optimization**:
   - While Rust's iterator methods are idiomatic, for performance-sensitive applications (especially when working with large arrays), traditional `for` loops may offer slight performance benefits due to lower overhead.
   - The refactored functions use `for` loops instead of iterator methods for both `manhattan_distance` and `hamming_distance`, reducing allocation and iterator overhead.

3. **Code Example** (Refactored `manhattan_distance`):
```rust
use anyhow::{Result, anyhow};

pub fn manhattan_distance(a: &[f32], b: &[f32]) -> Result<f32> {
    if a.is_empty() || b.is_empty() {
        return Err(anyhow!("Input arrays must not be empty"));
    }
    
    if a.len() != b.len() {
        return Err(anyhow!("Mismatched lengths: {} vs {}", a.len(), b.len()));
    }

    let mut sum = 0.0;
    for i in 0..a.len() {
        sum += (a[i] - b[i]).abs();
    }

    Ok(sum)
}
```

#### Why the Change?:

- **Performance Considerations**: Traditional for loops avoid the overhead that comes with iterators, which involve additional allocations and method calls. In some performance-critical situations, such as when dealing with large datasets, this can result in measurable improvements. Though the performance difference is often negligible, this change ensures that we are optimizing where necessary [0].

Simple Benchmark Test Rust Playground

- **Readability**: for loops make the flow of calculations more explicit and may be easier to follow for developers who are more familiar with traditional control structures.

#### Source:
![[0] Help, iterators made my Rust program slower! https://ntietz.com/blog/rust-iterators-and-threads/](https://ntietz.com/blog/rust-iterators-and-threads/)


arilotter:
Thanks for the PR! 
Do you have any benchmarks showing that the explicit for loops are faster than the iterator versions? we try not to prematurely optimize are fans of the iterator interface unless there's a good reason not to use it.

LVivona:
I ran benchmarks comparing the explicit for loop implementation against the iterator-based approach. Here are the results for a dataset of 10 million, as well as 10 thousand elements:

```rust
use criterion::{black_box, criterion_group, criterion_main, Criterion};
use psyche_core::{manhattan_distance, DistanceError};


fn manhattan_distance_iter(a: &[f32], b: &[f32]) -> Result<f32, DistanceError> {
    if a.is_empty() {
        return Err(DistanceError::EmptyInput);
    }
    
    if a.len() != b.len() {
        return Err(DistanceError::LengthMismatch(a.len(), b.len()));
    }
    
    Ok(a.iter().zip(b.iter()).map(|(x, y)| (x - y).abs()).sum())
}


pub fn manhattan_distance_benchmark(c: &mut Criterion) {
    let x = vec![1.0; 10000000];
    let y = vec![2.0; 10000000];

    c.bench_function("iterator", |b| {
        b.iter(|| black_box(manhattan_distance_iter(&x, &y)))
    });

    c.bench_function("for loop", |b| {
        b.iter(|| black_box(manhattan_distance(&x, &y)))
    });
}

// Register benchmarks
criterion_group!(benches, manhattan_distance_benchmark);
criterion_main!(benches);


```

### Sample 1000 Elements

```
hamming iterator        time:   [113.51 ns 113.97 ns 114.50 ns]
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) high mild
  3 (3.00%) high severe

hamming for loop        time:   [49.462 ns 49.568 ns 49.680 ns]
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

manhattan iterator      time:   [404.97 ns 405.29 ns 405.69 ns]
Found 11 outliers among 100 measurements (11.00%)
  4 (4.00%) high mild
  7 (7.00%) high severe

manhattan for loop      time:   [405.03 ns 405.53 ns 406.12 ns]
Found 10 outliers among 100 measurements (10.00%)
  3 (3.00%) high mild
  7 (7.00%) high severe
```

### Sample 10 Million Elements

**Note** ignore the change section.

```
hamming iterator        time:   [1.1653 ms 1.1677 ms 1.1701 ms]
                        change: [+170006145% +173386515% +176768717%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low severe
  2 (2.00%) low mild
  3 (3.00%) high mild
  2 (2.00%) high severe

hamming for loop        time:   [661.11 µs 662.03 µs 662.91 µs]
                        change: [+74162459% +75423116% +76805010%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  3 (3.00%) low mild
  1 (1.00%) high mild
  4 (4.00%) high severe

manhattan iterator      time:   [4.8736 ms 4.8806 ms 4.8893 ms]
                        change: [+1197050% +1201103% +1204449%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 9 outliers among 100 measurements (9.00%)
  3 (3.00%) high mild
  6 (6.00%) high severe

manhattan for loop      time:   [4.8715 ms 4.8768 ms 4.8835 ms]
                        change: [+1195008% +1199319% +1202737%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) high mild
  7 (7.00%) high severe
```
### Key Takeaways:
- For Manhattan Distance: The performance difference between the for loop and iterator is negligible (~0.05%), making both implementations viable.
- For Hamming Distance: The for loop is significantly faster, especially for large datasets.
- The iterator version introduces slightly more outliers, which may indicate minor fluctuations in execution time.

PR #50, opened 2025-05-15
fix: `tick_waiting_for_members` function

### LINKED ISSUE #51: bug: repeated checkpoint logic in`tick_waiting_for_members` function & incorrect unhealthy clients kicking logic

entropidelic:
Some modifications were done to this function at some time and some legacy and redundant code was left. This PR removes the redundant code and also fixes some lines that could bring a bug in some particular situations.

Basically, there was some program logic to check if we should change the checkpoint to `Hub`; this should happen if there are no peers from the previous epoch joining in the new one. This logic was repeated two times, so it was removed.
Also, unhealthy clients were kicked *after* the logic for changing checkpoint, so in some very edge cases the checkpoint could have not correctly set to `Hub` when it should.

Closes #51


IAvecilla (approved):
LGTM!

arilotter (approved):

PR #60, opened 2025-05-20
Fix fund_accounts.sh usage function calls 

nickgannon10:


arilotter:
can you tell us why should we merge this change, and what the reasoning behind it is?

"fix undefined function" on its own is not a clear enough commit message with no additional explainer.

PR #71, opened 2025-05-23
fix: reconnect to the gossip network if we have no peers

arilotter:
previously, we only checked our connection to the gossip network when we got a new state from the coordinator.
this meant that a network disconnection in the Warmup step would not be repaired until Warmup ended.

to fix this, we simply check our gossip connections every 10s & re-bootstrap as much as needed if we're offline.

### DIFF
--- a/shared/client/src/client.rs
+++ b/shared/client/src/client.rs
@@ -55,6 +55,7 @@ const REBROADCAST_SHAREABLE: Duration = Duration::from_secs(2);
 const DOWNLOAD_RETRY_BACKOFF_BASE: Duration = Duration::from_secs(2);
 const DOWNLOAD_RETRY_CHECK_INTERVAL: Duration = Duration::from_secs(1);
 const OPPROTUNISTIC_WITNESS_INTERVAL: Duration = Duration::from_millis(500);
+const CHECK_CONNECTION_INTERVAL: Duration = Duration::from_secs(10);
 
 impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static, B: Backend<T> + 'static>
     Client<T, A, B>
@@ -125,6 +126,7 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static, B: Backend<T> + 'sta
                 let mut sharing_downloadable_interval = interval(REBROADCAST_SHAREABLE);
                 let mut retry_check_interval = interval(DOWNLOAD_RETRY_CHECK_INTERVAL);
                 let mut opprotunistic_witness_interval = interval(OPPROTUNISTIC_WITNESS_INTERVAL);
+                let mut check_connection_interval = interval(CHECK_CONNECTION_INTERVAL);
                 let mut wait_for_checkpoint = false;
                 debug!("Starting client loop");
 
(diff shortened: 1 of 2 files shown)


jquesnelle:
@arilotter we had that other PR that got lost that only connected to peers if we actually were selected in the run -- does this new code for the p2p connection include that?

arilotter:
> @arilotter we had that other PR that got lost that only connected to peers if we actually were selected in the run -- does this new code for the p2p connection include that?

I believe that's https://github.com/PsycheFoundation/psyche/pull/64 - we should get both of these in. 64 first probably

shared/client/src/client.rs
```diff
@@ -596,6 +578,51 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static, B: Backend<T> + 'sta
...
+    if gossip_neighbors.is_empty() {
+        warn!("Not connected to any gossip peers! Trying to connect to some...");
+    }
+
+    const MAX_NUM_BOOTSTRAP_PEERS: usize = 3;
+    // we only want to bootstrap gossip;
+    // only connect to enough peers to bring our total peer count to at MOST MAX_NUM_BOOTSTRAP_PEERS.
+    // if we already have that many or more, don't send any gossip joins
+    // because gossip joins this way can force-disconnect other peers.
+    let num_peers_to_add = MAX_NUM_BOOTSTRAP_PEERS.saturating_sub(gossip_neighbors.len());
```
Frando, line 608: you can early-return here if num_peers_to_add == 0

shared/client/src/client.rs
```diff
@@ -596,6 +578,51 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static, B: Backend<T> + 'sta
     }
 }
 
+fn ensure_gossip_connected<T: NodeIdentity>(run_state: &Coordinator<T>, p2p: &mut NC) {
```
Frando, line 581: In principle this looks correct!

One note, as you're calling this fairly often (every 10 seconds): Gossip does currently not expose the number of "pending neighbor requests" - likely that should be the number you actually want to compare against (not only active neighbors). In happy situations it won't matter because 10s is plenty of time to establish a connection - unless nodes are stalled or unresponsive, then it takes 30s for connections to time out (unless you configured differently in the endpoint's transport config). 

If it happens that you re-enter `ensure_gossip_connected` while outgoing neighbor requests are still pending, it would lead to some of the previously-queued neighbors to be terminated once the now-queued neighbor connections succeed. This should not be an issue, but causes a bit more churn in the swarm. To prevent this, you could track the pending neighbor requests with the time they were pushed to gossip, and remove from the pending list once the NeighborUp event appears, and maybe log something & remove them after an interval. 

With that said: For the actual issue at hand (gossip losing all neighbors and not reconnecting) this PR should help without the additional tracking. (And iroh-gossip should expose enough information about pending connections so that users don't have to do that tracking themselves: https://github.com/n0-computer/iroh-gossip/issues/71)

ethernet8023: thanks - will leave as-is for now, left a comment to re-investigate later.

PR #76, opened 2025-05-27
feat: add authorization check and blob timeouts

jquesnelle:


arilotter (approved):
LGTM. the iroh-blobs fork also LGTM. This should fix a majority of our blob timeout/disconnection issues.

PR #82, opened 2025-05-28
Update idle timeout and retry configs

IAvecilla:
Instead of adding blob-level timeouts in iroh, we tweaked some configuration parameters to ensure that blob downloads time out at the connection level if no progress is made. These new settings apply a 5 second idle timeout and abort each failed download attempt immediately, so we can trigger our own retry logic rather than relying on iroh’s built-in retries.


entropidelic (approved):
lgtm

PR #90, opened 2025-05-30
Pythonify + Trainer abstraction phase 1

jquesnelle:
This is the first PR for adding Python to Psyche. There is now a `python` feature that will build in `pyo3`, and the entire project can be built as a Python extension and loaded into a virutalenv by running `maturin develop --release`

### DIFF
--- a/shared/coordinator/src/model.rs
+++ b/shared/coordinator/src/model.rs
@@ -33,11 +33,13 @@ unsafe impl ZeroableInOption for Model {}
     Deserialize,
     InitSpace,
     TS,
+    PartialEq,
 )]
 #[repr(C)]
 pub enum LLMArchitecture {
     HfLlama,
     HfDeepseek,
+    HfAuto,
 }
 
 #[derive(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2762,6 +2762,7 @@ dependencies = [
  "psyche-client",
  "psyche-modeling",
  "psyche-network",
+ "psyche-python-extension-impl",
  "tch",
 ]
 
@@ -6579,9 +6580,9 @@ checksum = "8b870d8c151b6f2fb93e84a13146138f05d02ed11c7e7c54f8826aaaf7c9f184"
 
 [[package]]
 name = "pkarr"
-version = "3.7.2"
+version = "3.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f552a2c2f80b6f3415e1722032ea92f24cc3f83f71b5b9de5600121c4a49fdd"
+checksum = "41a50f65a2b97031863fbdff2f085ba832360b4bef3106d1fcff9ab5bf4063fe"
 dependencies = [
  "async-compat",
  "base32",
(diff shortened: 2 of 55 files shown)


jquesnelle:
On the standalone extension side right now the basic building and tensor interop works

```
Python 3.10.12 (main, Jul 29 2024, 16:56:48) [GCC 11.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import psyche
>>> import torch
>>> x = torch.Tensor([1.])
>>> psyche._ext.add_one(x)
tensor([2.])
```

I've begun the trainer abstraction work, starting with at least getting inference to work. There's now `PythonCausalLM` which implements `CausalLM`. Right now I'm using `transformers` to load the model.

Native/no parallelism (`cargo run --release  --example inference -- --temperature 0`)
```
RICHARD:
I'll do the like, and wear them in my coat.

EDWARD:
I'll do the like, and wear them in my heart.

RICHARD:
I'll do the like, and wear them in my thoughts.

EDWARD:
I'll do the like, and wear them in my actions.
```

Native/TP 4 (`cargo run --release --features=parallelism  --example inference -- --temperature 0 --tensor-parallelism 4`)
```
RICHARD:
I'll do the like, and wear them in my coat.

EDWARD:
I'll do the like, and wear them in my heart.

RICHARD:
I'll do the like, and wear them in my thoughts.

EDWARD:
I'll do the like, and wear them in my actions.
```

Python/no parallelism (`cargo run --release --features=python  --example inference -- --temperature 0 --python)
```
RICHARD:
I'll do the like, and wear them in my coat.

EDWARD:
I'll do the like, and wear them in my cap.

RICHARD:
I'll do the like, and wear them in my heart.

EDWARD:
I'll do the like, and wear them in my mind.
```

Interesting that it's very slightly different, perhaps this means we won't be able to allow mix/match impls?

shared/coordinator/src/model.rs
```diff
@@ -33,11 +33,13 @@ unsafe impl ZeroableInOption for Model {}
     Deserialize,
     InitSpace,
     TS,
+    PartialEq,
 )]
 #[repr(C)]
 pub enum LLMArchitecture {
     HfLlama,
     HfDeepseek,
+    HfAuto,
```
jquesnelle, line 42: @crypto-vincent can you confirm that just adding an enum value here is not a breaking storage change? I would assume not but going from boolean-valued to more makes me slightly concerned

crypto-vincent: Yep, adding a case to a field-less `enum` should not break the memory layout.

An exception where would be if the `enum` had fields like so:

```rust
enum MyValueV1 {
  CaseA(u8),
}
enum MyValueV2 {
  CaseA(u8),
  CaseB(u64), // Adding this case would enlarge the enum's max size and this would break memory layout
}
```

Also reordering would break the memory layout if the case is added in the middle:

```rust
enum MyValueV1 {
   CaseA,
}
enum MyValueV2 {
   CaseB, // This would break the compatibility, since CaseB would steal CaseA's binary value (by ordering)
   CaseA,
}
```

PR #91, opened 2025-05-30
Use nix docker tooling to build images

IAvecilla:
This PR includes:
- Build all the docker images both for production and testing using nix dockerTools. 
- Update the docker docs
- Update the justfile commands adding a new nix module to improve the organization
- Remove all the dockerfiles


justfile
```diff
@@ -126,16 +126,31 @@ build_docker_test_client:
...
-    docker load < result
+    # nix build .#psyche-solana-test-validator
+    # docker load < result
     nix build .#psyche-solana-test-client-docker
     docker load < result
     cd docker/test && NUM_REPLICAS={{ num_clients }} docker compose up psyche-test-client -d --force-recreate
 
+run_docker_client:
+    nix build .#psyche-solana-client-docker
+    docker load < result
```
ethernet8023: if we switch from `pkgs.dockerTools.buildImage` to `pkgs.dockerTools.streamLayeredImage`, then instead of 
```
nix build .#psyche-solana-client-docker
docker load < result
```

we could do
```
nix run .#psyche-solana-client-docker | docker load
```

nix/packages.nix
```diff
@@ -94,54 +110,61 @@
...
               pathsToLink = [ "/bin" ];
             };
 
             config = {
+              Env = [ "NVIDIA_DRIVER_CAPABILITIES=all" ];
               Entrypoint = [ "/bin/client_test_entrypoint.sh" ];
             };
           };
 
           psyche-solana-test-validator = pkgs.dockerTools.buildImage {
```
ethernet8023: do we need a from Debian here? does it not run without one?

nix/packages.nix
```diff
@@ -94,54 +110,61 @@
...
-                    chmod +x $out/bin/psyche_solana_validator_entrypoint.sh
+                  mkdir -p $out/bin
+                  mkdir -p $out/local
+                  chmod 755 $out/local
+                  cp ${../docker/test/psyche_solana_validator_entrypoint.sh} $out/bin/psyche_solana_validator_entrypoint.sh
+                  cp -r ${coordinatorSrc} $out/local
+                  cp -r ${authorizerSrc} $out/local
+                  mv $out/local/*solana-coordinator $out/local/solana-coordinator
+                  mv $out/local/*solana-authorizer $out/local/solana-authorizer
+                  chmod +x $out/bin/psyche_solana_validator_entrypoint.sh
```
ethernet8023: i'm curious as to why we actually need to copy the src in here - what is it used for?

nix/packages.nix
```diff
@@ -94,54 +110,61 @@
             copyToRoot = pkgs.buildEnv {
               name = "root";
               paths = [
-                pkgs.solana-cli
+                inputs'.solana-pkgs.packages.solana
                 pkgs.bashInteractive
-                rustPackages."psyche-solana-client"
+                nixglhostRustPackages."psyche-solana-client-nixglhost"
                 (pkgs.runCommand "entrypoint" { } ''
```
ethernet8023: this feels.. strange to me, but i can't put my finger on why. :thinking:

arilotter (approved):

PR #93, opened 2025-06-02
WIP Dynamic Batch Assignment: simpler version

dsocolobsky:


shared/client/src/state/train.rs
```diff
@@ -141,7 +141,7 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static> TrainingStepMetadata
     pub fn start(
         &mut self,
         client_index: u64,
-        state: &Coordinator<T>,
+        state: &mut Coordinator<T>,
```
entropidelic: I think this shouldn't be a mutable reference. Also, double check the other functions where the coordinator is now changed to a mutable reference, since I think they shouldn't be either.
As a rule of thumb, we should pay special attention when changing a coordinator reference to mutable in a the client code, since this would mean every client could change the coordinator's state, and this could add hidden bugs

shared/client/src/state/witness.rs
```diff
@@ -111,11 +114,60 @@ impl WitnessStep {
...
+        };
+
+        let step = state.progress.step - 1; // First step is 1, so we subtract 1 to get 0-index
+
+        // Cycle through the partitions depending on the current step, based on actual client count.
+        let slice_index = if num_actual_slices == 0 {
+            0 // Should never happen but just in case
+        } else {
+            step % (num_actual_slices as u32)
+        };
```
entropidelic: is there any reason why we are calculating and sending this index in the witness? 
I think this value should be centrally calculated, tracked and updated by the coordinator in each step, and witness nodes should get it through the coordinator state

shared/coordinator/src/coordinator.rs
```diff
@@ -965,39 +997,106 @@ impl<T: NodeIdentity> Coordinator<T> {
...
+            for i in removed_clients_idx.iter() {
+                agreed_values.remove(*i);
+            }
+            msg!(
+                "[tick_round_witness] Agreed time values: {:?}",
+                agreed_values
+            );
+
+            // Here we assume the offset is the same for all of the witnesses, we should run a consensus on that.
+            let start_offset_in_global_times = witnesses[0].training_times.offset;
```
entropidelic: just as another argument for centrally managing the training times offset, the 0th witness could have lied with the offset sent and this would lead to errors in this code

shared/coordinator/src/coordinator.rs
```diff
@@ -965,39 +997,106 @@ impl<T: NodeIdentity> Coordinator<T> {
...
-            // disconnected. We just set everyone to withdrawn state and change
-            // to Cooldown.
-            if num_witnesses == 0 {
-                self.withdraw_all()?;
-                self.start_cooldown(unix_timestamp);
-                return Ok(TickResult::Ticked);
+        let witnesses = &self.current_round_unchecked().witnesses;
+        let mut client_times_last_index_updated = 0;
+        if let Some(mut agreed_values) =
+            self.get_consensus_on_reported_times(witnesses, 1.0 / 3.0, 5.0)
```
entropidelic: we should make the parameters for the threshold and the tolerance constants somewhere in the code, so that they are more visible and easily configurable

shared/client/src/state/train.rs
dsocolobsky: yes good catch, it was probably left over from before some of the new changes

shared/coordinator/src/coordinator.rs
dsocolobsky: true, although I think ideally they should work dynamically based on a % of the actual values but I couldn't find a simple and efficient way to do that

shared/coordinator/src/coordinator.rs
```diff
@@ -965,39 +997,106 @@ impl<T: NodeIdentity> Coordinator<T> {
...
+        if (client_times_last_index_updated + 1) >= self.epoch_state.clients.len() {
+            msg!("[tick_round_witness] Persisting assigned batch sizes for all clients");
+            let assigned_batch_sizes = self.calculate_batch_sizes_per_client();
+            for i in 0..assigned_batch_sizes.len() {
+                let client = self.epoch_state.clients.get_mut(i);
+
+                if let Some(client) = client {
+                    client.assigned_batch_size = assigned_batch_sizes.get(i).cloned().unwrap_or(1);
+                }
             }
```
entropidelic, line 1097:
```suggestion
            for (client_idx, client) in self.epoch_state.clients.iter_mut().enumerate() {
                client.assigned_batch_size =
                    assigned_batch_sizes.get(client_idx).cloned().unwrap_or(1);
            }
```

entropidelic: also I don't really like doing the `.unwrap_or(1)` here, I think it could hide some error in the code, since the `get()` call should never return `None` in this context

shared/client/src/state/witness.rs
dsocolobsky: hmm yes I can't remember why I did it like this, it's probably better to keep track of it on the coordinator I'll make the change

shared/coordinator/src/data_selection.rs
```diff
@@ -10,20 +11,22 @@ pub fn assign_data_for_state<T: NodeIdentity>(
 ) -> BTreeMap<BatchId, T> {
     let round = coordinator.current_round().unwrap();
 
-    let trainer_nodes: Vec<_> = (0..coordinator.epoch_state.clients.len())
+    // Get trainer nodes references to avoid cloning
+    let mut trainer_nodes: Vec<_> = (0..coordinator.epoch_state.clients.len())
         .filter_map(|i| {
-            let client = &coordinator.epoch_state.clients[i];
+            let client = coordinator.epoch_state.clients.get(i)?;
```
entropidelic: small nit, I think you can iterate directly over the clients and indicies with enumerate

```rust
coordinator.epoch_state.clients.iter().enumerate().filter_map(|(i, client)| ...)
```

shared/coordinator/src/coordinator.rs
dsocolobsky: Yes, if we want correctness then it should always be present, however for resilience we might want to not crash and just assign 1 for now and in the next rounds it will eventually get fixed. We can discuss it.

shared/client/src/state/witness.rs
dsocolobsky: ok should be changed now in the latest commit!

shared/coordinator/src/coordinator.rs
dsocolobsky: Left it like that but added error logs

shared/client/src/state/train.rs
```diff
@@ -160,28 +160,32 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static> TrainingStepMetadata
         let cancel_training = CancellationToken::new();
         let round_start = Instant::now();
 
-        let round = state.current_round().ok_or(TrainError::NoActiveRound)?;
+        // Extract some values in order to avoid the borrow checker
+        let (round_height, round_random_seed, round_tie_breaker_tasks) = {
+            let r = state.current_round().ok_or(TrainError::NoActiveRound)?;
+            (r.height, r.random_seed, r.tie_breaker_tasks)
+        };
```
IAvecilla, line 167: I don't know if it's more idiomatic but you can do:

```rust
let (round_height, round_random_seed, round_tie_breaker_tasks) = state
   .current_round()
   .ok_or(TrainError::NoActiveRound)
   .and_then(|r| Ok((r.height, r.random_seed, r.tie_breaker_tasks)))?;
```

Feel free to choose

shared/client/src/state/witness.rs
```diff
@@ -111,11 +112,43 @@ impl WitnessStep {
...
+            let source_idx = client_index_start + i;
+
+            if source_idx < clients_len {
+                let _ = training_times.push(
+                    current_round
+                        .client_times
+                        .get(source_idx)
+                        .copied()
+                        .unwrap_or(0),
+                );
```
IAvecilla: Should we break from the loop if `source_idx > clients_len`? I think we are not doing anything once we reach that point

shared/coordinator/src/data_selection.rs
```diff
@@ -32,60 +38,68 @@ pub fn assign_data_for_state<T: NodeIdentity>(
...
-        if node_batch_size > 0 {
-            let end_index = current_index + node_batch_size - 1;
-            assignments.insert(
-                BatchId(ClosedInterval::new(current_index, end_index)),
-                node.id,
-            );
-            current_index = end_index + 1;
-        }
+    // Use references in the loop to avoid moving/copying data
+    for (client, node_id) in &trainer_nodes {
```
IAvecilla: This should do it
```suggestion
    for (client, node_id) in trainer_nodes.iter() {
```

shared/coordinator/src/data_selection.rs
```diff
@@ -32,60 +38,68 @@ pub fn assign_data_for_state<T: NodeIdentity>(
...
     coordinator: &Coordinator<T>,
-    num_trainer_nodes: u64,
+    committee_selection: &CommitteeSelection,
 ) -> Vec<BatchId> {
-    let start = round.data_index;
-    let total_size = coordinator.get_target_global_batch_size(Some(round)) as u64;
-    let end = start + total_size;
+    // Get the list of trainer nodes and their original indices, similar to assign_data_for_state
+    // The elements are (original_client_index, reference_to_client_data)
+    let mut trainer_info: Vec<_> = (0..coordinator.epoch_state.clients.len())
```
IAvecilla: I think we're doing exactly the same above, we should make a function in the coordinator to get the trainer nodes and mimic the other logic (seems nicer)

shared/coordinator/src/data_selection.rs
```diff
@@ -32,60 +38,68 @@ pub fn assign_data_for_state<T: NodeIdentity>(
...
-    let mut batch_ids = Vec::with_capacity(num_trainer_nodes as usize);
-    let mut current = start;
+    let mut batch_ids = Vec::with_capacity(trainer_info.len());
+    let mut current_data_idx = round.data_index;
 
-    for i in 0..num_trainer_nodes {
-        let node_size = base_size + if i < remainder { 1 } else { 0 };
+    for (client_original_idx, _client_ref) in &trainer_info {
+        let client = coordinator.epoch_state.clients.get(*client_original_idx);
+        let Some(client) = client else {
```
IAvecilla: Maybe we can do `if let Some(client) = client {...}` and include the main logic there instead of checking for the contrary first

shared/coordinator/src/data_selection.rs
```diff
@@ -32,60 +38,68 @@ pub fn assign_data_for_state<T: NodeIdentity>(
...
+    deterministic_shuffle(&mut trainer_info, round.random_seed);
 
-    let mut batch_ids = Vec::with_capacity(num_trainer_nodes as usize);
-    let mut current = start;
+    let mut batch_ids = Vec::with_capacity(trainer_info.len());
+    let mut current_data_idx = round.data_index;
 
-    for i in 0..num_trainer_nodes {
-        let node_size = base_size + if i < remainder { 1 } else { 0 };
+    for (client_original_idx, _client_ref) in &trainer_info {
```
IAvecilla:
```suggestion
    for (client_original_idx, _client_ref) in trainer_info.iter() {
```

shared/coordinator/src/data_selection.rs
```diff
@@ -32,60 +38,68 @@ pub fn assign_data_for_state<T: NodeIdentity>(
...
+            *node_id,
+        );
+        current_index = end_index + 1;
     }
 
+    msg!("[assign_data_for_state] assignments: {:?}", assignments);
     assignments
 }
 
 pub fn get_batch_ids_for_round<T: NodeIdentity>(
```
IAvecilla, line 29: It seems that the function here and `assign_data_for_state` are almost identical, except that one sets the assignments and the other sets the batch IDs. Maybe we could create a single function that takes an enum or something similar to specify the desired data, like:

```rust
enum RoundData {
     Assignments,
     BatchIDs,
}
 ```

or something along those lines. A new struct might also be worth considering, but we can discuss it further.

File: shared/coordinator/src/coordinator.rs [hunk 14]
```diff
@@ -1122,6 +1242,177 @@ impl<T: NodeIdentity> Coordinator<T> {
...
+
+        // For each time slot
+        for i in 0..TRAINING_TIMES_SLICE_SIZE {
+            let mut values: Vec<u16> = witnesses
+                .iter()
+                .map(|w| w.training_times.get(i).copied().unwrap_or(0))
+                .collect();
+
+            // If all values are 0, that's the consensus
+            if values.iter().all(|&v| v == 0) {
```
IAvecilla: Does this mean that the witnesses didn't receive any time for this client? Maybe we can do an enum like:

```rust
enum TrainingTimes {
     NotReceived,
     Received(u16)
}
```

And work with that, it will be better than checking 0 as the wrong value. I don't know if this will bring any problem to the witness message but we can think about it.

shared/coordinator/src/coordinator.rs
```diff
@@ -1122,6 +1242,177 @@ impl<T: NodeIdentity> Coordinator<T> {
...
+                agreed_values.push(0);
+                continue;
+            }
+
+            // Avoid 0s as they should be invalid reported times
+            values.retain(|&x| x != 0);
+            if values.is_empty() {
```
IAvecilla, line 1271: I don't think we ever reach this case, if all values are zero we are already doing the continue above. Also I'm not sure why here we return `None` to say there's no consensus but we agree on the value in the check above.

IAvecilla (commented):
A bunch of nits and improvement suggestions, nothing critical

shared/coordinator/src/coordinator.rs
dsocolobsky: I will check but I think I tried with Option<u16> and the problem is that it's not zeroable so I couldn't add it to the struct, this enum looks like it might cause the same.

shared/client/src/state/train.rs
dsocolobsky: looks a bit more complex to me, I'll stay with the current one for now but thanks!

shared/coordinator/src/data_selection.rs
dsocolobsky: Since this is a `Vec` the .iter() is implicit so no need for that, if it were a `FixedVec` then yeah I think we need to do `.iter()`.

The comment is old and was removed, thanks!

shared/coordinator/src/coordinator.rs
dsocolobsky: yes, well spotted, we only need one check and we should continue not return None, as the consensus is for that particular client and we should keep checking the rest

shared/client/src/state/witness.rs
dsocolobsky: yes good idea

shared/coordinator/src/data_selection.rs
dsocolobsky: done!

shared/coordinator/src/data_selection.rs
dsocolobsky: usually I prefer early return but in this case it looks shorter so done!

shared/coordinator/src/data_selection.rs
dsocolobsky: ok I managed to do a bit of a refactor in ca528d7e3c5667ce6383c7ee2aba120050b4e04a but not sure, seems to work though

IAvecilla: I like it!

shared/coordinator/src/data_selection.rs
dsocolobsky: same I dont think its necesary as Rust adds it implicitly in a `Vec`

shared/coordinator/src/coordinator.rs
```diff
@@ -90,6 +98,8 @@ pub struct Client<I> {
     pub id: I,
     pub state: ClientState,
     pub exited_height: u32,
+    pub round_train_time: u16,
+    pub assigned_batch_size: u16,
```
crypto-vincent, line 104: nit: this is probably a breaking change to the coordinator's layout, would be worth checking if the test introduced in this PR still passes:

dsocolobsky: Yeah it makes it fail
```
called `Result::unwrap()` on an `Err` value: IncorrectSize { expected: 145696, actual: 119072 }
```

Which I guess was to be expected, perhaps we can hold on a few breaking PRs and figure out a single migration later?

crypto-vincent: Yeah I would think so too, Jeff had some plans for that I think

PR #94, opened 2025-06-02
Fix model evaluations spikes

pefontana:
We are seeing some spikes in the model evaluations.
image
We believe it is because clients are reporting benchmark scores with few samples, causing spikes in the evaluation graph.
Now, clients will need to wait until their buffer is full to report model scores.


arilotter (approved):
lgtm, makes sense, frontend will deal with this fine.

what do you think about also adding a random offset, so we don't start at the beginning of the benchmark every time?

PR #97, opened 2025-06-04
DO NOT MERGE - Parallelize build weighted index

ryleung-solana:
Experiments to try to parallelize or at least speed up build_weighted_index

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7199,6 +7199,7 @@ dependencies = [
  "psyche-watcher",
  "rand 0.8.5",
  "rand_chacha 0.3.1",
+ "rayon",
  "reqwest 0.12.15",
  "serde",
  "serde_json",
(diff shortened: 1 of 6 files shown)


ryleung-solana:
This implementation seems to run out of stack space running the parallel shuffle on real data sizes, as the parallel sort and shuffle are divide-and-conquer algorithms, that use rayon::join to wrap the recursive calls, which, due to work-stealing will eventually just causes the same worker threads to run the recursive calls. I did have the idea to basically re-implement the rayon parallel algorithms in tokio and return futures for the recursive calls, and have an outer "driving loop" await the futures for each "level", effectively turning the depth-first algorithm breadth-first. However, this would involve re-implementing and maintaining the parallel algorithms in tokio, which given that this algorithm is complicated enough as it is, with many potential edge cases, and requiring extensive validation, seems not worth it at the moment in terms of back/buck.

PR #109, opened 2025-06-12
fix: hunting down remaining Repr(Rust)

crypto-vincent:
We use bytemuck for serialization/deserialization, meaning that we use the Raw RAM layout for serialization. The RAM layout depends on the `Repr()`.

One big No-no here is that the `Repr(Rust)` does NOT guarantee anything layout wise and its actual behaviour is undefined (although compiler internals as of today make it somewhat predictable for simple cases).

This means that `Repr(Rust)` is BY DEFINITION not cross-platform (aka typescript incompatible for example)

This PR hunts down the last few remaining forgotten `Repr(Rust)` but with most importantly: THIS PR DOES NOT CHANGE THE BINARY LAYOUT, just translate to the non `Repr(Rust)` current equivalent of the current behaviour.

### DIFF
--- a/shared/coordinator/src/coordinator.rs
+++ b/shared/coordinator/src/coordinator.rs
@@ -22,7 +22,6 @@ pub const WAITING_FOR_MEMBERS_EXTRA_SECONDS: u64 = 3;
 // bloom filter with 1024 bits (16 u64)
 pub type WitnessBloom = Bloom<16, 8>;
 
-#[repr(u8)]
 #[derive(
     Clone,
     Copy,
@@ -37,6 +36,7 @@ pub type WitnessBloom = Bloom<16, 8>;
     InitSpace,
     TS,
 )]
+#[repr(u8)]
 pub enum RunState {
     #[default]
     Uninitialized = 0,
@@ -49,7 +49,6 @@ pub enum RunState {
     Paused = 7,
 }
 
-#[repr(u8)]
 #[derive(
     Clone,
     Copy,
@@ -64,6 +63,7 @@ pub enum RunState {
     InitSpace,
     TS,
 )]
+#[repr(u8)]
 pub enum ClientState {
     #[default]
     Healthy = 0,
@@ -85,6 +85,7 @@ pub enum ClientState {
     TS,
 )]
 #[serde(bound = "I: NodeIdentity")]
+#[repr(C)]
 pub struct Client<I> {
     pub id: I,
     pub state: ClientState,
@@ -202,6 +203,7 @@ impl WitnessEvalResult {
 }
 
 #[derive(Clone, Copy, Debug)]
+#[repr(u8)]
 pub enum CoordinatorError {
     NoActiveRound,
     InvalidWitness,
--- a/shared/core/src/fixed_string.rs
+++ b/shared/core/src/fixed_string.rs
@@ -27,6 +27,7 @@ pub struct FixedStringError((usize, usize));
     InitSpace,
     Zeroable,
 )]
+#[repr(C)]
 pub struct FixedString<const L: usize>(
     #[serde(
         serialize_with = "serde_serialize_string",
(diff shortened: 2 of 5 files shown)


shared/coordinator/src/coordinator.rs
```diff
@@ -37,6 +36,7 @@ pub type WitnessBloom = Bloom<16, 8>;
     InitSpace,
     TS,
 )]
+#[repr(u8)]
```
crypto-vincent, line 39: this is just moved for clarity close to the actual name

shared/coordinator/src/coordinator.rs
```diff
@@ -64,6 +63,7 @@ pub enum RunState {
...
```
crypto-vincent, line 66: this is just moved for clarity close to the actual name

shared/coordinator/src/coordinator.rs
```diff
@@ -85,6 +85,7 @@ pub enum ClientState {
     TS,
 )]
 #[serde(bound = "I: NodeIdentity")]
+#[repr(C)]
```
crypto-vincent, line 88: This is an actual change (as the `Repr(Rust)` was the default here)

However and this is important: both `Repr(Rust)` and `Rust(C)` have the exact same binary outcome here, by pure luck. So it is NOT a breaking change, it just correctly instructs the IDL to use the `Repr(C)` instead of the default `Repr(Rust)` of which the behaviour is undefined and compiled dependent.

shared/coordinator/src/coordinator.rs
```diff
@@ -202,6 +203,7 @@ impl WitnessEvalResult {
 }
 
 #[derive(Clone, Copy, Debug)]
+#[repr(u8)]
```
crypto-vincent, line 206: Same here, just clarifing the behaviour, the default behaviour is the same as `Repr(u8)` for fieldless enums in `Repr(rust)`

shared/core/src/fixed_string.rs
```diff
@@ -27,6 +27,7 @@ pub struct FixedStringError((usize, usize));
     InitSpace,
     Zeroable,
 )]
+#[repr(C)]
```
crypto-vincent, line 30: Same as the above, this used to be a `Repr(Rust)` but by sheer luck this is binary equivalent to `Repr(C)` in the current rust implementation, switching to `Repr(C)` to ensure the layout doesn't change when we switch rust compiler version

shared/core/src/fixed_vec.rs
```diff
@@ -7,6 +7,7 @@ use ts_rs::TS;
 
 #[derive(Clone, Copy, Zeroable, AnchorSerialize, AnchorDeserialize, TS)]
 #[ts(type = "Array<T>", bound = "T: TS")]
+#[repr(C)]
```
crypto-vincent, line 10: Same as the above, this used to be a `Repr(Rust)` but by sheer luck this is binary equivalent to `Repr(C)` in the current rust implementation, switching to `Repr(C)` to ensure the layout doesn't change when we switch rust compiler version

shared/coordinator/src/coordinator.rs
crypto-vincent: to clarify: Switching to `Repr(C)` to ensure the layout doesn't change when we switch rust compiler version.

crypto-vincent:
> all look good to me. holy moly we got lucky, eh?
> 
> is there a way we can enforce that everything in the coordinator is repr(c), i wonder?

I believe there is no way to enforce that we use `Repr(c)` all the way down since for example `Option<>` is a `Repr(Rust)` but there's nothing we can do about that (other than making our `COption<>`.

It just so happens that we're lucky on several ends because the rust compiler behaviour is relatively simple for simple cases:
- 1) Enums that are fieldless and `Repr(Rust)` behave exactly like `Repr(u8)`, but that's not guarantee, just an accidental feature of current rust version
- 2) Structs with less than 3 fields does NOT get field re-ordered by the current rust compiler version (so `Client<>` for example was lucky about this)
- 3) Enums with less than 2 fields behave predictably on this rust compiler version (`Option<>` for example, so the IDL can properly process those)

PR #111, opened 2025-06-12
Add Prompt Task for Text Generation

pefontana:
### Overview

This PR introduces the ability for the client to perform a text generation task (a `PromptTask`) in addition to running model evaluations. This new functionality is activated by passing the `--prompt-task` flag to the `client train` command.

### Implementation Details

- To handle both evaluation and prompt tasks uniformly, this PR introduces the `EnumModelTask` enum:

```
  pub enum EnumModelTask {
      EvalTask(EvalTask),
      PromptTask(PromptTask),
  }
```

* The client executes different logic depending on the variant of this enum.
* The PromptTask generates text one token at a time. These generated tokens are then sent in the WitnessMetadata.
* Currently, the initial prompt passed to the LLM is selected randomly from a hardcoded list.

### Tx info vs MAX_TOKENS_TO_SEND
* MAX_TOKENS_TO_SEND= 8 
 Program log: Instruction: Witness
Program HR8RN2TP9E9zsi2kjhvPbirJWA1R6L6ruf4xNNGpjU5Y consumed 39731 of 200000 compute units
Tx size: 1034 bytes

* MAX_TOKENS_TO_SEND= 16
Program log: Instruction: Witness
Program HR8RN2TP9E9zsi2kjhvPbirJWA1R6L6ruf4xNNGpjU5Y consumed 39806 of 200000 compute units
Tx size: 1066 bytes

* MAX_TOKENS_TO_SEND= 48
Program log: Instruction: Witness
Program HR8RN2TP9E9zsi2kjhvPbirJWA1R6L6ruf4xNNGpjU5Y consumed 40996 of 200000 compute units
Tx size: 1194 bytes

* MAX_TOKENS_TO_SEND= 50
Program log: Instruction: Witness
Program HR8RN2TP9E9zsi2kjhvPbirJWA1R6L6ruf4xNNGpjU5Y consumed 41026 of 200000 compute units
Tx size: 1202 bytes

#60 Fail

Max transaction size is 1232 bytes

### Next Steps
With the introduction of `EnumModelTask`, some variable names could be clearer, and the code for prompt and evaluation tasks could be better separated.
A follow-up PR will address this with a small refactor to improve code clarity.


shared/client/src/state/prompt.rs
```diff
@@ -0,0 +1,109 @@
...
+        // sample next token
+        let mut logits_processor =
+            LogitsProcessor::from_sampling(rand::random(), Sampling::All { temperature: 0.6 });
+
+        let next_token = match logits_processor.sample(&logits) {
+            Ok(token) => token,
+            Err(_) => {
+                panic!("Failed to sample next token");
+            }
+        };
```
dsocolobsky: nit but if you're going to panic isn't this the same as

```rust
let next_token = logits_processor.sample(&logits)
    .expect("Failed to sample next token");
```

shared/client/src/state/prompt.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[derive(Debug)]
+pub struct PromptTask {
+    _selected_promt: usize,
+    tokens: RwLock<Vec<i32>>,
+    pub tokens_to_send: RwLock<FixedVec<i32, TOKENS_TO_SEND_LENGTH>>,
+    /// A flag set to `true` once the end-of-sequence token has been generated.
+    pub prompt_finished: RwLock<bool>,
+    /// next_index exists primarily to maintain a compatible interface with
+    /// evaluation tasks. Here only track progress the progress of the prompt.
```
ethernet8023: typo - progress the progress

shared/client/src/state/prompt.rs
```diff
@@ -0,0 +1,109 @@
...
+use tch::Tensor;
+use tokenizers::Tokenizer;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, trace};
+
+const MAX_CONTEXT_LENGTH: usize = 2048;
+
+#[derive(Debug)]
+pub struct PromptTask {
+    _selected_promt: usize,
```
ethernet8023: typo - should be `_selected_prompt`

shared/client/src/state/prompt.rs
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        let tokens = self.tokens.read().unwrap();
+
+        let input = Tensor::from_slice(&tokens)
+            .to(*trainer.device())
+            .unsqueeze(0);
+
+        // drop tokens to release lock
+        drop(tokens);
```
ethernet8023: rather than an explicit drop here, what if you did
```rs
// the tokens.read() lock is released at the end of this block
let input = {
    let tokens = self.tokens.read().unwrap();
    Tensor::from_slice(&tokens)
       .to(*trainer.device())
       .unsqueeze(0)
};
```
to release the lock implicitly at the end of the scope?

shared/client/src/state/prompt.rs
```diff
@@ -0,0 +1,109 @@
...
+        // Run forward pass
+        let (logits, _) = trainer.forward(&input, None, Some(1));
+
+        let logits = logits.squeeze();
+
+        // sample next token
+        let mut logits_processor =
+            LogitsProcessor::from_sampling(rand::random(), Sampling::All { temperature: 0.6 });
```
ethernet8023, line 76: for reproducibility, maybe we should do a lower temperature?
more curiously - why 0.6?

shared/client/src/state/prompt_texts.rs
```diff
@@ -0,0 +1,82 @@
+pub const PROMPT_TEXTS_LEN: usize = 2;
```
ethernet8023: rather than storing PROMPT_TEXTS_LEN separately, maybe we just `PROMPT_TEXTS_LEN.len()` everywhere? i know they're enforced to be in sync here, but feels a lil smelly to have them separate

shared/client/src/state/prompt_texts.rs
```diff
@@ -0,0 +1,82 @@
...
+
+pub const PROMPT_TEXTS: [&str; PROMPT_TEXTS_LEN] = [
```
ethernet8023: maybe we could load these from a folder using something like `include_str!` - would make reading them from the website much easier

shared/client/src/state/stats.rs
```diff
@@ -236,21 +244,48 @@ impl StatsLogger {
...
+    pub fn get_prompt_results(&self) -> FixedVec<i32, TOKENS_TO_SEND_LENGTH> {
+        let mut results = FixedVec::new();
+        for eval_task in self.eval_runner.tasks().iter().flatten() {
+            if eval_task.name() == PROMPT_TASK_NAME {
+                match &eval_task.task {
+                    EnumModelTask::PromptTask(prompt_task) => {
+                        let tokens = prompt_task.tokens_to_send.read().unwrap();
+                        results.extend(tokens.iter().cloned()).unwrap();
+                        // Release lock
+                        drop(tokens);
```
ethernet8023: maybe just put these two lines in a block? explicit drop always feels a bit weird to me

shared/client/src/state/stats.rs
```diff
@@ -236,21 +244,48 @@ impl StatsLogger {
...
+                        prompt_task.tokens_to_send.write().unwrap().clear();
+                    }
+                    _ => tracing::warn!("Unexpected eval task type"),
```
ethernet8023: maybe warn the type in the message here, if possible?

shared/coordinator/src/coordinator.rs
```diff
@@ -18,6 +18,8 @@ pub const SOLANA_MAX_NUM_WITNESSES: usize = 32;
 pub const BLOOM_FALSE_RATE: f64 = 0.01f64;
 pub const WITNESS_QUORUM_RAIO: f64 = 2.0f64 / 3.0f64;
 pub const WAITING_FOR_MEMBERS_EXTRA_SECONDS: u64 = 10;
+// max amount of tokens to send in a witness message
+pub const TOKENS_TO_SEND_LENGTH: usize = 8;
```
ethernet8023: i worry that 8 tokens won't be super exciting to see. it's possible we have to go out-of-band for this - is >8 not possible / too expensive?

arilotter:
one more nit - the commit `update solana-pkgs.url =` does not accurately reflect the contents of the commit

shared/client/src/state/prompt.rs
pefontana: Sounds reasonable to reduce it. I just used 0.6 because it is the default one in the inference example
I modified it to 0.3, What do you think?

shared/client/src/state/prompt_texts.rs
pefontana: You are right!
IDK why I made it that way hahah

arilotter (approved):

arilotter:
TODO:
- >  *Once the prompt reaches the end of sequence, or the task has generated a large number of tokens, we should start again with a new prompt. (Right now we just set the prompt task to finished and stop there)

we should never stop, it should simply loop.

- test the real prompts on some checkpoints
- add the real prompts

PR #116, opened 2025-06-16
fix: epoch rewards now being properly distributed

crypto-vincent:


jquesnelle:
@crypto-vincent just confirming my read that this is on-chain logic only and should be a transparent upgrade

crypto-vincent:
> @crypto-vincent just confirming my read that this is on-chain logic only and should be a transparent upgrade

Yes that's correct 👌 no memory layout change, just the behaviour of the tick at the end of the epoch is changed here. Should be fairly safe, should be also fully transparently upgradeable.

shared/coordinator/src/coordinator.rs
```diff
@@ -439,28 +439,18 @@ impl<T: NodeIdentity> Coordinator<T> {
         unix_timestamp: u64,
         random_seed: u64,
     ) -> std::result::Result<TickResult, CoordinatorError> {
-        let ret = match self.run_state {
+        match self.run_state {
```
crypto-vincent, line 442: note: this change should have the exact same behaviour as before, just reducing code. Not sure why the logic was extra convoluted here (probably because of historical changes?)

architectures/decentralized/solana-tooling/src/process_coordinator_instructions.rs
```diff
@@ -163,6 +164,34 @@ pub async fn process_coordinator_set_paused(
         .await
 }
 
+pub async fn process_coordiantor_set_future_epoch_rates(
```
dsocolobsky, line 167: nit but there's a typo here and in the other uses of this function

should be `process_coordinator_set_future_epoch_rates`

crypto-vincent: Woops! good point ahah, will fix in a different PR since this is only in tests 👌

jquesnelle:
Had to revert, was access violating at cooldown

https://explorer.solana.com/tx/5QyoZSSsA5EHGwvcvUySq6zDVBMXTnpLaubsw9Z4YYNHPkxqJRJLzyMrNEeswm15v9GJTmyTZrVqpa9nNzhVcqdk?cluster=devnet

PR #122, opened 2025-06-17
Make `WAITING_FOR_MEMBERS_EXTRA_TIME` a config value

### LINKED ISSUE #105: refactor: move `WAITING_FOR_MEMBERS_EXTRA_SECONDS` to a config variable
In the main Coordindator struct there's a hardcoded constant `WAITING_FOR_MEMBERS_EXTRA_SECONDS` which sets how long we wait until after our `min_clients` have been met to allow for additional people to join in (so we don't just start with the minimum). This should be a configuration variable, not hardcoded.

NOTE: This would be a breaking change to the contract storage

IAvecilla:
Close #105.

### DIFF
--- a/shared/coordinator/src/coordinator.rs
+++ b/shared/coordinator/src/coordinator.rs
@@ -17,7 +17,6 @@ pub const SOLANA_MAX_NUM_WITNESSES: usize = 32;
 
 pub const BLOOM_FALSE_RATE: f64 = 0.01f64;
 pub const WITNESS_QUORUM_RAIO: f64 = 2.0f64 / 3.0f64;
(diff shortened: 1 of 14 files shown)


shared/coordinator/src/coordinator.rs
```diff
@@ -251,6 +250,7 @@ pub struct CoordinatorConfig {
     pub global_batch_size_end: u16,
 
     pub verification_percent: u8,
+    pub waiting_for_members_extra_time: u8,
```
crypto-vincent, line 259: just a nit but i'm a bit worried this introduces a breaking change, could you check if the new test added by this PR still passes?

PR #123, opened 2025-06-18
Handle many data providers

### LINKED ISSUE #47: Handle training data server dying
- You should be able to run multiple mirrors of a data provider
- Whether Huggingface / online / from the data provider server, Witnesses should be able to attest that the data provider is dead. if a consensus of witnesses agrees that the data provider is dead, the run should be suspended & nobody penalized for not training.

@dsocolobsky:
> I've started working on this, I thought of allowing the config to define several data providers and then if one fails it can check the next, etc. as backup , dunno if that's what you meant.

> 2nd point is good too but would that mean that if one of the options for data providers dies then we suspend the whole run? or do we want backups?

I think multiple providers for the same data is perfect, yeah, just have to do a sanity check that they actually have similar data / same length at startup when creating a run at least.
Fallback would be great, but a retry for http would be great first imo

dsocolobsky:
This PR was opened in the old repo, so I'm moving it here.

* Adds redundancy in data providers, you can now specify more than one in the config file as backups and they will be used in case the first one defined does not work.
* In the case of an error when downloading from a data provider; first it will attempt retry a few times, afterwards it will switch to the next defined data provider. Next time it downloads it will remember the last working data provider.
* Overall the logic changes are not too complex but this Introduces breaking changes to the config file formats
* You can also now define "failing" Dummy data providers which are just used for testing and will always return error, used to test this.
* Adds a new decentralized integration test which tests the main functionality in this PR.

Closes #47 though it does not implement witnessing for non-available data providers, I think that might be better off in a separate PR after testing this first instance is working, this adds a bit more resilience with little complexity.


shared/coordinator/src/model.rs
```diff
@@ -181,17 +203,22 @@ pub struct LLM {
     pub architecture: LLMArchitecture,
     pub checkpoint: Checkpoint,
     pub data_type: LLMTrainingDataType,
-    pub data_location: LLMTrainingDataLocation,
+    pub data_locations: FixedVec<LLMTrainingDataLocation, { MAX_DATA_LOCATIONS }>,
```
crypto-vincent, line 209: nit: this might be a breaking change in the coordinator's onchain layout, would be worth making sure the test introduced in this PR still passes:

dsocolobsky: yep it fails

```
called `Result::unwrap()` on an `Err` value: IncorrectSize { expected: 119992, actual: 119072 }
```

PR #124, opened 2025-06-20
docs: breaking changes

crypto-vincent:


ryleung-solana:
I would personally add as a criterion to decide between centralized and decentralized approaches to migrations, CU considerations - this is why e.g. program deploys are usually done over many txs, and the main advantage of the centralized approach is that we can do the writes over many txs to build up that buffer. As I understand it (correct me if I'm wrong) the main reason we currently have our current custom deserializer and lack convenient backward compatibility is to save CUs due to the large size of the state, which would seem to preclude deserializing and reserializing the whole state like the centralized approach would need...

ryleung-solana:
Yeah, for sure, I think the best scenario is we basically figure out a new serialized data format that's backward compatible and less monolithic then do 1 migration and be done with it.

crypto-vincent:
> Yeah, for sure, I think the best scenario is we basically figure out a new serialized data format that's backward compatible and less monolithic then do 1 migration and be done with it.

Yeah, the good news is that if the current data on-chain is not critical (aka we already saved the weights of the current training somewhere else) we could probably just re-deploy a new run-id with a different program-id and run with it, but time will tell what we'll go for.

crypto-vincent:
> I pushed a fix for the the CI failure due to the `fixtures` dir not being included in the flake. Looks like there's still another test fail now, though

amazing, looks like everything is good to go now 👌

jquesnelle (approved):

PR #127, opened 2025-06-21
fix: retry fixing reward distribution

crypto-vincent:
## Summary

Giving another shot to fixing the reward distribution logic from the revert PR: https://github.com/PsycheFoundation/psyche/pull/116

## Details

Retrying by not using a `HashSet` during reward distribution but instead keeping original client ordering when initiating the epoch's client list (so that the reward distribution can just do a sequential scan on the client list)

### DIFF
--- a/architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/lib.rs
+++ b/architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/lib.rs
@@ -270,7 +270,7 @@ pub struct OwnerCoordinatorAccounts<'info> {
         bump = coordinator_instance.bump,
         constraint = coordinator_instance.main_authority == authority.key()
     )]
-    pub coordinator_instance: Account<'info, CoordinatorInstance>,
+    pub coordinator_instance: Box<Account<'info, CoordinatorInstance>>,
 
     #[account(
         mut,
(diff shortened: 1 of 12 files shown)


architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/lib.rs
```diff
@@ -270,7 +270,7 @@ pub struct OwnerCoordinatorAccounts<'info> {
         bump = coordinator_instance.bump,
         constraint = coordinator_instance.main_authority == authority.key()
     )]
-    pub coordinator_instance: Account<'info, CoordinatorInstance>,
+    pub coordinator_instance: Box<Account<'info, CoordinatorInstance>>,
```
crypto-vincent, line 273: making sure to put the `CoordinatorInstance` on the heap to not crowd the stack

shared/coordinator/src/coordinator.rs
```diff
@@ -896,16 +886,22 @@ impl<T: NodeIdentity> Coordinator<T> {
             let height = self.current_round_unchecked().height;
             self.move_clients_to_exited(height);
 
+            // Read the pending clients
+            let mut pending_clients_ordered = Vec::with_capacity(pending_clients.len());
+            let mut pending_clients_unordered = HashSet::with_capacity(pending_clients.len());
```
jquesnelle, line 891: I guess we just have more compute room here to do the HashSet?

crypto-vincent: Yep, HashSet in itself doesn't use much stack so it's fine in this case since there's some space in that particular execution path. (it's the other execution path where the stack is almost full before we even instanciate the HashSet)

jquesnelle (approved):

jquesnelle:
@crypto-vincent can you fix up the branch?

Issue #98, opened 2025-06-06
bug: wandb errors can be fatal


jquesnelle:
Got an error where wandb failed to create a run

```
Error: Error applying state: Failed to init run in warmup: wandb failed to create run: api request failed: Request error:
HTTP status client error (409 Conflict) for url (https://api.wandb.ai/graphql)
Response body:
{"errors":[{"message":"Error 1062 (23000): <REDACTED>}
```

This propagates all the way up and crashes the process. Probably want to just gracefully fail to log to wandb (but keep on training)

AaronBryant:
Is this bug because the weights and biases are unknown? or it can't figure out exactly which ones to use? 

I'm trying to start my own country out of the Midwest (it's a generational project) but I'm having trouble figuring out exactly what it means to be midwestern. The borders I've discovered are the Mississippi river, the Ohio river, and the great lakes. Should I accept only those born / raised in the midwest or welcome outsiders who live here now? What about new people?

PR #132, opened 2025-06-25
Do not panic on wandb failure

dsocolobsky:
The application will continue and we log an error to the console, in the future we might want to retry the operating in a separate thread in the background or similar, but for now this is the simplest solution.

Originally we tried to do the retry logic but we had to made some functions async and it got messy really quickly, we can take a look at it in the future if we deem it important.

Solves #98


jquesnelle (approved):

AaronBryant:
Did anyone get paid for fixing this? I got an email about it being fixed. @dsocolobsky If you want, I will add you on LinkedIn.

PR #140, opened 2025-06-27
Share peer state for retry and model sharing logic

IAvecilla:
This PR introduces changes to the P2P model sharing logic by restructuring it around two new actor-style components: `PeerManagerHandle` and `RetriedDownloadsHandle`. Each spawns an internal actor that manages peer state or retried download tracking through dedicated channels. This approach removes any shared locking and ensures serialized, centralized handling of peer and retry state, preventing main loop locks and race conditions.

Additionally, one test was removed as it no longer aligns with the updated architecture. However, the underlying logic it verified is still covered by existing tests in the `network` module.

### DIFF
--- a/shared/client/src/state/evals.rs
+++ b/shared/client/src/state/evals.rs
@@ -93,7 +93,7 @@ impl EvalRunner {
                 eval_tasks
                     .into_iter()
                     .map(|task| {
-                        let prepared = task.prepare(&tokenizer, eval_task_max_docs);
+                        let prepared = task.prepare(&tokenizer, None, eval_task_max_docs);
                         Arc::new(EvalTask {
                             task: prepared,
                             results: Arc::new(RunningAverage::new()),
--- a/shared/eval/examples/evaluate.rs
+++ b/shared/eval/examples/evaluate.rs
@@ -49,7 +49,7 @@ fn main() -> Result<()> {
     )?;
     for task in tasks {
         let name = format!("{task}");
-        let result = task.prepare(&tokenizer, None).run(
+        let result = task.prepare(&tokenizer, None, None).run(
             EvalTaskOptions {
                 model: model.as_mut(),
                 skip_and_step_by: None,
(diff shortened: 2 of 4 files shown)


pefontana (approved):

shared/client/src/client.rs
```diff
@@ -785,38 +781,29 @@ fn all_node_addrs_shuffled<T: NodeIdentity>(state: &Coordinator<T>) -> Vec<NodeA
...
-    let router = p2p.router();
-    // initialize variables to request model config
-    let config_blob_tickets = Arc::new(std::sync::Mutex::new(Vec::with_capacity(1)));
-    let errored_peers = Arc::new(std::sync::Mutex::new(HashMap::new()));
-    let num_peers = peer_ids.len();
-    let peer_cycle = Arc::new(Mutex::new(VecDeque::from(peer_ids)));
-    if num_peers == 0 {
-        return Err(anyhow::anyhow!("No peers available to request the model"));
-    }
+    let blob_ticket = Arc::new(std::sync::Mutex::new(Vec::with_capacity(1)));
```
pefontana, line 789: Why this need to be an Arc mutex? 
Only so we can use the fn param_request_task?

architectures/decentralized/testing/tests/integration_tests.rs
```diff
@@ -244,100 +244,6 @@ async fn test_rejoining_client_delay() {
     }
 }
 
-#[test_log::test(tokio::test(flavor = "multi_thread"))]
-#[serial]
-async fn test_client_removed_from_p2p_sharing_model_list_after_failing_many_times() {
```
pefontana, line 249: We remove it because the is no sharing model list now?

shared/client/src/client.rs
IAvecilla: Yes, exactly the `param_request_task` is also used in the parameter sharing logic and there we spawn tasks for each parameter and get the blob tickets from that vec so it needs to be an arc mutex for each task to push the parameter and we can read it from the outside

architectures/decentralized/testing/tests/integration_tests.rs
IAvecilla: Yes something like that, like checking the peer list that is sharing now is not part of the client logic and instead has its own actor so we don't have particular access to the list in the integration tests. I'll try to have a test replacing this same behavior

jquesnelle (approved):

PR #151, opened 2025-07-04
nix: housekeeping

isabelroses:
hopefully some good work on that nix clean up you wanted @arilotter :D

i did my best to also split up the commits.

quite a bit of my work really ended up being on cohesion of the nix rather then the packages themselves. they are quite well managed by crane :P

### DIFF
--- a/nix/lib.nix
+++ b/nix/lib.nix
@@ -1,8 +1,9 @@
 {
-  gitcommit,
-  inputs,
-  system,
   pkgs,
+  inputs,
+  lib ? pkgs.lib,
+  gitcommit ? inputs.self.rev or inputs.self.dirtyRev or "unknown",
+  system ? pkgs.stdenv.hostPlatform.system,
 }:
 let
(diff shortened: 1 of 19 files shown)


nix/lib.nix
```diff
@@ -145,15 +151,15 @@ let
...
+    pkgs.runCommandNoCC "${package.name}-nixgl-wrapped"
       {
         nativeBuildInputs = [ pkgs.makeWrapper ];
       }
       ''
         mkdir -p $out/bin
         for bin in ${package}/bin/*; do
           if [ -f "$bin" ] && [ -x "$bin" ]; then
-            makeWrapper "$bin" "$out/bin/$(basename $bin)" \
+            wrapProgram "$bin" \
```
ethernet8023: this won't work, because wrapProgram tries to `mv` the original executable while `makeWrapper` will leave the original in-place - and here, we don't have the permission to move the original.

PR #152, opened 2025-07-05
fix: Zellic treasurer audit pre-draft recommendations

crypto-vincent:


architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/logic/participant_claim.rs
```diff
@@ -78,15 +75,16 @@ pub fn participant_claim_processor(
     {
         if client.id.signer == context.accounts.user.key() {
             participant_earned_points = client.earned;
+            break;
```
crypto-vincent, line 78: original feedback:
---------


Also in participant_claim, we have a for loop that iterates over clients
```rust
for client in context
    .accounts
    .coordinator_account
    .load()?
    .state
    .clients_state
    .clients
    .iter()
{
    if client.id.signer == context.accounts.user.key() {
        participant_earned_points = client.earned;
    }
}
```
the clients are a Vec that can have up to 256 entries. It is unclear whether clients is a list of unique entries or if there can be multiple duplicate clients.
In the first case, this for loop is unnecessarily iterating over all 256 possible entries, even though we might have already found our match at position 0 thus wasting a lot of CU.
In the latter case, participant_earned_points is overwritten with the latest found entry of client.earned and not added together. The fix for this would depend on the intended logic.

architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/logic/participant_claim.rs
```diff
@@ -78,15 +75,16 @@ pub fn participant_claim_processor(
...
 
     let participant = &mut context.accounts.participant;
     let run = &mut context.accounts.run;
 
-    if params.claim_earned_points
-        > participant_earned_points - participant.claimed_earned_points
-    {
+    let participant_unclaimed_points = participant_earned_points
+        .saturating_sub(participant.claimed_earned_points);
+    if params.claim_earned_points > participant_unclaimed_points {
```
crypto-vincent: original feedback:
------------

In participant_claim we have some math operations that are not using safe math and could potentially under/overflow. For example:

```rust
if params.claim_earned_points
    > participant_earned_points - participant.claimed_earned_points
{
    return err!(ProgramError::InvalidParameter);
}
```

architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/logic/participant_claim.rs
```diff
@@ -12,9 +12,6 @@ use crate::ProgramError;
 #[derive(Accounts)]
 #[instruction(params: ParticipantClaimParams)]
 pub struct ParticipantClaimAccounts<'info> {
-    #[account(mut)]
-    pub payer: Signer<'info>,
-
```
crypto-vincent, line 17: original feedback:
----------

In two instructions, namely run_top_up and participant_claim a payer (that has to be a signer) is a required account, however that account is never used in the function.

```rust
#[account(mut)]
pub payer: Signer<'info>,
```

architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/logic/participant_claim.rs
```diff
@@ -78,26 +75,40 @@ pub fn participant_claim_processor(
...
-    participant.claimed_collateral_amount += claim_collateral_amount;
-    participant.claimed_earned_points += params.claim_earned_points;
-
-    run.total_claimed_collateral_amount += claim_collateral_amount;
-    run.total_claimed_earned_points += params.claim_earned_points;
+    participant.claimed_collateral_amount = participant
+        .claimed_collateral_amount
+        .checked_add(claim_collateral_amount)
+        .ok_or(error!(ProgramError::MathOverflow))?;
+    participant.claimed_earned_points = participant
```
crypto-vincent: Note: this is a bit verbose, since we already compile with:
```
overflow-checks = true
```
this might be overkill, but i'll let the team decide 👌

architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/logic/participant_claim.rs
crypto-vincent: Note: we use the `overflow-checks = true` compilation flag to mitigate this issue

architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/logic/participant_claim.rs
ethernet8023: Maybe we should switch our impl to some set-like instead of a vec-like, later.

PR #155, opened 2025-07-08
feat: remove unecessary top-up and improve tests

crypto-vincent:
## Summary

Optional change to simplify some of the logic of the treasurer.

I noticed that the "top_up" instruction is relatively useless, it only increments a "funded_collateral_amount" counter which itself is already a duplicate of the "run_collateral.amount"

## Details

This PR simply remove the "funded_collateral_amount" field and the "top_up" IX

Removing this IX reduce the complexity of the program this PR also includes more tests to increase reliability

### DIFF
--- a/psyche-book/src/enduser/client-faq.md
+++ b/psyche-book/src/enduser/client-faq.md
@@ -1,8 +1,8 @@
 # Client FAQ
 
 - Which operating systems are supported?
-  - We support officially suport modern Linux versions, with Mac support planned for the future once the Metal backend is implemented.
-- Which are the hardware requirements to run a client?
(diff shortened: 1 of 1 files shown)


architectures/decentralized/solana-treasurer/programs/solana-treasurer/src/state/run.rs
```diff
@@ -14,7 +14,6 @@ pub struct Run {
 
     pub collateral_mint: Pubkey,
 
-    pub total_funded_collateral_amount: u64,
```
crypto-vincent, line 17: breaking change, but the contract is not yet deployed/used on-chain yet so it's fine

PR #161, opened 2025-07-12
feat: add treasurer to CLI

crypto-vincent:
## Summary

Add support for treasurer instructions from the CLI used for the raw coordinator


architectures/decentralized/solana-client/src/backend.rs
```diff
@@ -290,122 +290,39 @@ impl SolanaBackend {
...
-            .signer(coordinator_account_signer.clone())
+            .signer(coordinator_account_signer)
             .send()
             .await?;
 
-        let mut create_signatures = vec![create_coordinator_signature];
-
-        if join_authority == payer {
-            let (authorization_create, authorization_activate) =
-                self.create_run_ensure_permissionless().await?;
```
crypto-vincent, line 322: I believe we've got `scripts/create_authorization..sh` for this now

architectures/decentralized/solana-client/src/backend.rs
```diff
@@ -290,122 +290,39 @@ impl SolanaBackend {
...
-                        params: psyche_solana_coordinator::logic::InitCoordinatorParams {
-                            main_authority,
-                            join_authority,
-                            run_id,
-                        },
-                    })
-                    .instructions()
-                    .unwrap()[0]
-                    .clone(),
+                } else {
```
crypto-vincent, line 305: this is the core changes: we used to directly call the coordinator instruction, but now it's a conditional action:

```
if (we_are_using_treasurer) {
   // call treasurer instruction
} else {
   // call coordiantor directly as before
}

File: architectures/decentralized/solana-client/src/main.rs [hunk 3]
```diff
@@ -341,9 +359,27 @@ async fn async_main() -> Result<()> {
                 CommitmentConfig::confirmed(),
             )
             .unwrap();
+
+            if treasurer_index.is_some() && treasurer_collateral_mint.is_none() {
```
jquesnelle, line 363: How does one determine their `treasurer_index`?

crypto-vincent: We either hand-pick it by specifying it through the flag or we let the CLI auto-generate one for us from the run_id's hash. TLDR: we can pick a value and stick to it (or the CLI will pick a value for us in the background if we dont)

PR #163, opened 2025-07-14
feat: add eval questions inference

entropidelic:


dsocolobsky:
To test you can run something like this to run 3 prompts:

```
cargo run --release --example inference -- --tasks "boolq" --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --revision main --seed 32 --temperature 0.1 --eval-index 115 --eval-limit 3
```

You can also add `--json` for json-formatted output for easy parsing.

shared/modeling/examples/inference.rs
```diff
@@ -95,22 +97,39 @@ struct Args {
...
 
 fn inference(
     repo_files: Vec<PathBuf>,
     tensor_parallelism: Option<(CommunicatorId, usize, usize, Arc<Barrier>)>,
     args: Args,
-    seed: u64,
     mut tokens: Vec<i64>,
     tokenizer: Tokenizer,
-) -> Result<()> {
+    print_to_stdout: bool,
```
entropidelic: considering we are passing the `args` input to the function and you can take this value from there, this may be a little redundant

dsocolobsky: good catch I didn't see that

PR #164, opened 2025-07-14
fix nix setup on macos

arilotter:
also fix a safari rendering bug. thanks safari.

this doesn't get the python stuff building on MacOS, but the devshell and formatter work at least.

this should also massively decrease first evaluation times for the devshell on other platforms, by not building all the workspace deps.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -311,16 +311,15 @@
         ]
       },
       "locked": {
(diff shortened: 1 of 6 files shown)


crypto-vincent (approved):
it works on my M3 Max 🥳 nice!

PR #165, opened 2025-07-16
Extra logs for P2P model sharing process

IAvecilla:
This PR updates some logs to use `info` level, mostly around the P2P model sharing process, to improve observability. It also includes two small changes that simplify the logic without altering the core functionality:

* We no longer differentiate between errors during the blob request and those occurring mid-download. If a peer fails 5 times at any point in the process, it will be removed and not used again.
* The retry window for selecting an available peer to request a blob ticket has been reduced from 8 minutes to 4. If no peers become available within that time, the process will be canceled.

### DIFF
--- a/shared/client/src/client.rs
+++ b/shared/client/src/client.rs
@@ -1,24 +1,24 @@
 use crate::{
+    Broadcast, BroadcastType, ClientTUIState, Finished, IntegrationTestLogMarker, NC,
+    RunInitConfig, RunInitConfigAndIO, TrainingResult,
     state::{ApplyMessageOutcome, DistroBroadcastAndPayload, FinishedBroadcast, RunManager},
-    Broadcast, BroadcastType, ClientTUIState, Finished, IntegrationTestLogMarker, RunInitConfig,
-    RunInitConfigAndIO, TrainingResult, NC,
 };
-use anyhow::{bail, Error, Result};
+use anyhow::{Error, Result, bail};
 use futures::future::join_all;
 use psyche_coordinator::{Commitment, CommitteeSelection, Coordinator, RunState};
 use psyche_core::NodeIdentity;
 use psyche_metrics::{ClientMetrics, ClientRoleInRound, PeerConnection};
 use psyche_network::{
-    allowlist, blob_ticket_param_request_task, raw_p2p_verify, router::Router,
     AuthenticatableIdentity, BlobTicket, DownloadComplete, DownloadRetryInfo, DownloadType,
-    ModelRequestType, NetworkConnection, NetworkEvent, NetworkTUIState, Networkable, NodeAddr,
-    NodeId, PeerManagerHandle, RetriedDownloadsHandle, SharableModel, TransmittableDownload,
-    MAX_DOWNLOAD_RETRIES,
+    MAX_DOWNLOAD_RETRIES, ModelRequestType, NetworkConnection, NetworkEvent, NetworkTUIState,
+    Networkable, NodeAddr, NodeId, PeerManagerHandle, RetriedDownloadsHandle, SharableModel,
+    TransmittableDownload, allowlist, blob_ticket_param_request_task, raw_p2p_verify,
+    router::Router,
 };
 use psyche_watcher::{Backend, BackendWatcher};
 use tokenizers::Tokenizer;
 
-use rand::{seq::SliceRandom, thread_rng, RngCore};
+use rand::{RngCore, seq::SliceRandom, thread_rng};
 use std::{
     collections::{BTreeSet, HashMap},
     marker::PhantomData,
@@ -27,7 +27,7 @@ use std::{
 };
 use tokio::{
     select,
-    sync::{mpsc, watch, Notify},
+    sync::{Notify, mpsc, watch},
     task::JoinHandle,
     time::interval,
 };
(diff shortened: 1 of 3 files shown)


shared/network/src/lib.rs
```diff
@@ -652,12 +652,12 @@ fn parse_gossip_event<BroadcastMessage: Networkable>(
...
-                        message_hash = message_hash,
-                        "received gossip message with hash {message_hash}: {:?}",
-                        result
-                    );
+                    // debug!(
+                    //     name: "gossip_rx",
+                    //     message_hash = message_hash,
+                    //     "received gossip message with hash {message_hash}: {:?}",
+                    //     result
+                    // );
```
dsocolobsky: We can perhaps leave this as trace! and just not enable trace usually

shared/network/src/p2p_model_sharing.rs
```diff
@@ -91,123 +84,107 @@ impl PeerManagerHandle {
...
                 self.errors_per_peers = HashMap::from_iter(errors_per_peers_vec);
 
-                debug!(
+                info!(
                     "Updated peer list: {} peers available to ask for the model parameters",
                     self.available_peers.len()
                 );
             }
             PeerCommand::GetPeer { reply } => {
+                info!("AVAILABLE PEERS: {:?}", self.available_peers);
```
dsocolobsky:
```suggestion
                debug!("AVAILABLE PEERS: {:?}", self.available_peers);
```
I would leave this as debug since it's pretty verbose.

Or we could also figure out a way to shorten each pubkey for the clients, so for example show only the first 4 characters.

shared/network/src/p2p_model_sharing.rs
```diff
@@ -91,123 +84,107 @@ impl PeerManagerHandle {
...
                 };
                 let _ = reply.send(peer);
             }
             PeerCommand::ReportSuccess { peer_id } => {
-                self.available_peers.push_back(peer_id);
-                debug!("Peer {peer_id} correctly provided the blob ticket");
+                if !self.available_peers.contains(&peer_id) {
+                    self.available_peers.push_back(peer_id);
+                } else {
+                    error!("Peer was already available but we tried to add it again");
```
dsocolobsky:
```suggestion
                    warn!("Peer was already available but we tried to add it again");
```

I guess it's a warn! since it's not fatal

shared/network/src/p2p_model_sharing.rs
```diff
@@ -91,123 +84,107 @@ impl PeerManagerHandle {
...
-                            .all(|(_, (e, _))| *e == self.max_retries_per_peer)
-                    {
-                        error!(
-                            "No more peers available to download blob tickets, terminate process"
-                        );
-                        cancellation_token.cancel();
+                    if !self.available_peers.contains(&peer_id) {
+                        self.available_peers.push_back(peer_id);
+                    } else {
+                        error!("Peer was already available but we tried to add it again");
```
dsocolobsky:
```suggestion
                        warn!("Peer was already available but we tried to add it again");
```

Same

shared/client/src/client.rs
```diff
@@ -266,22 +267,28 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static, B: Backend<T> + 'sta
...
                                             TransmittableDownload::DistroResult(distro_result) => {
                                                 debug!("Download complete: step {} batch id {}", distro_result.step, distro_result.batch_id);
                                                 run.apply_distro_result(hash, distro_result, None);
                                             },
                                             TransmittableDownload::ModelParameter(parameter) => {
-                                                debug!("Download complete: parameter {}", parameter.name()?);
+                                                current_downloaded_parameters += 1;
+                                                info!("Download complete: parameter {}", parameter.name()?);
+                                                if let Some(total_parameters) = total_parameters {
+                                                    info!("Downloaded parameters total: {}/{}", current_downloaded_parameters, total_parameters);
```
entropidelic, line 281: nit
```suggestion
                                                    info!("Total downloaded parameters: {}/{}", current_downloaded_parameters, total_parameters);
```

shared/client/src/client.rs
```diff
@@ -266,22 +267,28 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static, B: Backend<T> + 'sta
...
+                                                } else {
+                                                    error!("Total parameters not set");
+                                                }
```
entropidelic: I will add a little comment explaining that this should never really happen, it just a safeguard log in case something really weird happens

ethernet8023: is it possible to continue correctly if total parameters isn't set? if not we should error

shared/network/src/lib.rs
IAvecilla: We commented out just for testing but I'll leave it as it is, I know it's a little verbose but it is useful to ensure that the peers are getting messages via gossip

shared/network/src/p2p_model_sharing.rs
IAvecilla: It was just for testing purposes and forgot to remove it, I’m not sure it gives us much useful information. We already have some logs for when the available peers list is empty, and one for when we successfully get a peer assigned, so I think that should be enough.

shared/client/src/client.rs
IAvecilla: It’s just a variable we set at the beginning of the param request process to help with logging each completed download. It shouldn’t happen, but if the variable isn’t set correctly while the actual list of parameters is still fine, then the process can continue, we just won’t be able to log the total number of parameters.

entropidelic (approved):

PR #171, opened 2025-07-21
Fix MMLU PRO: Add generate until task type

pefontana:
# Add generate until task type

* Add `TaskType::GenerateUntil`, a task type where the model generates text until it reaches a specific stopping condition (e.g., a special token, pattern, or maximum length)
* Modify MmluPro evaluation to use the GenerateUntil task instead of log likelihood task type
* Refactor the `min_samples` flow to distinguish between different task types, allowing each to have its own minimum sample threshold before reporting scores
* Hardcode fewshots amounts when the client runs the eval in training. MmluPro = 5 and 0 for the rest of the evaluations
* Remove acc_norm metric in evaluations that don't need it (boolq, mmlu and mmlu_pro)

## Benchmarks
Since MMluPro can take a very long time to run (even more than 24 hours with 1 GPU), I made a reduced dataset https://huggingface.co/datasets/pefontana/mmlu_pro_reduced with 5% of the original dataset, distinguished by category

**Model**: [DeepSeek-V2-Lite](https://huggingface.co/deepseek-ai/DeepSeek-V2-Lite)

* Psyche evals:  `MMLU Pro: {"acc": 0.24177631578947367}`
* lm_eval:  
```
    "mmlu_pro": {
      "exact_match,custom-extract": 0.22697368421052633,
      "exact_match_stderr,custom-extract": 0.016764641497841758,
      "alias": "mmlu_pro"
    },
```


shared/client/src/cli.rs
```diff
@@ -228,24 +225,22 @@ impl TrainArgs {
...
 
     pub fn eval_tasks_from_args(
         eval_tasks: &str,
-        eval_fewshot: usize,
         eval_seed: u64,
     ) -> Result<Vec<psyche_eval::Task>> {
         let result: Result<Vec<psyche_eval::Task>> = eval_tasks
             .split(",")
             .map(|eval_task| {
+                let fewshot = { if eval_task == "mmlu_pro" { 5 } else { 0 } };
```
entropidelic, line 241: if this argument does not depend on any input, maybe we can set it directly inside ther `Task::new()` function

Issue #174, opened 2025-07-22
web backend: support multiple coordinator contracts


arilotter:
we currently only index one.
we'll need to index runs from multiple coordinators as we grow & push breaking changes.
we should prefix our data with the program addr, so as we deploy different versions, we can support runs with the same ID but at different coordinators.

dsocolobsky:
Just for confirmation the idea is to support more than one run/coordinator and also distinguish between runs with the same name but with different program IDs in the coordinator?

<img width="829" height="898" alt="Image" src="https://github.com/user-attachments/assets/a1589867-0ef7-40ec-9507-0b532e24ca94" />

arilotter:
@dsocolobsky yeah - we already support more than one run, but only one coordinator program. if we deploy a second one, I'd like runs from both to be shown to the frontend, and just want to ensure that we can handle two runs with the same run id from two different coordinator programs.

dsocolobsky:
Thanks for the answer @arilotter. Is the mining pool the same for all coordinators or should we support multiple ones as well?

dsocolobsky:
By the way @arilotter in the PR I didn't add the mining pool changes yet, regarding that I'm not sure how to proceed:

1. We don't have an UI to support multiple mining pools no? Like the list of coordinators, or we don't need that and only one will show for each run?
2. A run supports just 1 mining pool or several? I assume a mining pool can be present in multiple coordinators.

PR #179, opened 2025-07-24
backend/frontend handle multiple coordinators

dsocolobsky:
Allows the website to track runs from multiple coordinators.

* Changes route `/runs/$run/$index` to `/runs/$run/$programId/$index`
* Changes runKeys from `runId-index` to `programId-runId-index`
* Sets backend to listen to `0.0.0.0:PORT` instead of `127.0.0.1:PORT`
* Includes nix fix for anchor from another branch
* Fixed shutdown on backend server (doing Ctrl+C used to hang and would never quit).

Multiple mining-pools is still TODO perhaps it's better in a separate PR? We'd need to think the UI for that I think.

solves #174

### DIFF
--- a/website/backend/src/index.ts
+++ b/website/backend/src/index.ts
@@ -111,16 +111,37 @@ async function main() {
 	const liveRunSummaryListeners: Set<(runData: RunSummariesData) => void> =
 		new Set()
 
-	coordinator.dataStore.eventEmitter.addListener('updateSummaries', () => {
-		const summaries = coordinator.dataStore.getRunSummaries()
-		for (const listener of liveRunSummaryListeners) {
-			try {
-				listener(summaries)
-			} catch (err) {
-				console.error(`Failed to send run summaries to subscribed client...`)
+	// Set up event listeners for run summaries from all coordinators
+	for (const [_programId, coordinator] of coordinators) {
+		coordinator.dataStore.eventEmitter.addListener('updateSummaries', () => {
+			// Aggregate summaries from all coordinators
+			let allRuns: any[] = []
+			let totalTokens = 0n
+			let totalTokensPerSecondActive = 0n
+
+			for (const [_, coord] of coordinators) {
+				const coordinatorSummary = coord.dataStore.getRunSummaries()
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -34,7 +34,7 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
     solana-pkgs = {
-      url = "github:arilotter/solana-flake";
+      url = "github:arilotter/solana-flake/anchor-weirdness";
```
IAvecilla: If this worked maybe we can ask Ari to merge it and just use the original solana-flake main branch. Just to avoid problems in the future with the versioning of the flake

website/backend/src/index.ts
```diff
@@ -274,10 +347,16 @@ async function main() {
 			res.send(stream)
 
 			function sendRunData(runData: RunData) {
+				// Calculate total runs across all coordinators
+				let totalRuns = 0
+				for (const [_, coordinator] of coordinators) {
+					totalRuns += coordinator.dataStore.getNumRuns()
+				}
+
```
IAvecilla, line 427: I think you're doing exactly the same thing above, maybe we can add a new function `get_total_runs` or something to retrieve this information

IAvecilla (commented):
Some nits but looks good!

website/backend/src/index.ts
dsocolobsky: true, solved

website/backend/src/chainTracker.ts
```diff
@@ -25,117 +25,195 @@ interface TimestampedError {
...
+	try {
+		const files = readdirSync(stateDirectory)
+		const coordinatorFiles = files.filter(
+			(file) => file.startsWith('coordinator-db-') && file.endsWith('.json')
+		)
+
+		return coordinatorFiles.map((file) => {
+			// Extract program ID from filename: coordinator-db-{programId}.json
+			const programId = file.slice('coordinator-db-'.length, -'.json'.length)
+			return programId
```
entropidelic: small nit: it could be sligthly easier to follow if you make the mapping first storing it in a variable and then returning it

PR #180, opened 2025-07-24
local telemetry, new metrics, and cardinality reduction

entropidelic:
The idea of this PR is to add a custom dashboard that can be useful as a starting point to monitor the state of each client more easily.

In the dashboard, there is a dropdown to select each of the connected clients:

Screenshot 2025-07-25 at 3 29 31 PM

Once a client is selected, you can see three visualization rows.
The first one, shows a couple of general metrics of the client state, like the current step, the number of gossip neighbors and a plot visualizing when the client finished training in a step:

Screenshot 2025-07-25 at 3 34 33 PM

The second one shows some visualizations for the p2p model sharing state of the client, mostly relevant when the client joins a run and requests the parameters to other peers:

Screenshot 2025-07-25 at 3 42 58 PM

The third one shows the logs for that client:

Screenshot 2025-07-25 at 3 46 42 PM

---

A readme has been also added for spawning the telemetry stack locally in `/telemetry/README.md`. This is only meant as a starting point, suggestions for new metrics or to make the ones present better are appreciated!

### DIFF
--- a/architectures/centralized/server/src/main.rs
+++ b/architectures/centralized/server/src/main.rs
@@ -172,7 +172,7 @@ async fn main() -> Result<()> {
                     MetricsDestination::OpenTelemetry(OpenTelemetry {
                         endpoint,
                         authorization_header: run_args.oltp_auth_header.clone(),
-                        report_interval: Duration::from_secs(10),
+                        report_interval: Duration::from_secs(60),
                     })
                 }))
                 .with_trace_destination(run_args.oltp_tracing_url.clone().map(|endpoint| {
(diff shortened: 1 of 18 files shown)


entropidelic:
> ty, this looks great
> 
> The biggest question I usually have is "for each client, how many blobs have they downloaded for the current round and the previous rounds" and "how many broadcast announces for the current round have they received". Can I see that with this dashboard? Basically if we have a node that drops off, I want to look at that client and see something like "Oh, he didn't finish downloading 7 of the blobs" or "he never saw the broadcasts from client Y" etc

those seem like some really good metrics to add. With a little extra work It should definetely be possible to include them to the dashboard

PR #185, opened 2025-07-29
feat: finetuning

jquesnelle:
This adds support for finetuning

 - `PreprocessedDataProvider` which reads a HuggingFace dataset produced by this script: https://github.com/NousResearch/torchtitan/blob/3b714320f4dbdb1d48f26fd973a1cb8ee0dfd3b7/scripts/preprocess_data.py
 - Add `position_ids` as model argument so samples can have their RoPE position reset when packing
 - Adds explicit `FlashAttention2` attention implementation that bypasses Torch's SDPA to call Flash Attention 2 directly. This is done by way of the `_flash_attention_forward`  LibTorch function, which gives access to the `varlen` version of FA2, giving us multi-sequence packing (FA2 does not support explicit attention masks so this is the only way to prevent cross-sample attention)


shared/data-provider/src/preprocessed.rs
```diff
@@ -0,0 +1,165 @@
...
+                                    inputs_column,
+                                    Some(num_tokens_per_sequence),
+                                )?;
+                                let labels = match labels_column {
+                                    Some(column) => Some(list_to_vec(
+                                        &row,
+                                        column,
+                                        Some(num_tokens_per_sequence),
+                                    )?),
+                                    None => None,
```
entropidelic, line 110: small rust nit: the `Option<T>` type has the `.map()` method, which applies a closure to the wrapped value when having a `Some` variant, and leaves the `None` unchanged, making it a little less verbose

```suggestion
                                    let labels = labels_columns.map(|column| 
                                       list_to_vec(
                                        &row,
                                        column,
                                        Some(num_tokens_per_sequence),
                                    )?)
```
a similar pattern can be used with `position_ids` and `sequence_lenghts`

PR #189, opened 2025-07-30
Add macOS/aarch64 development support

tahsintunan:
### Summary
This PR adds development environment support for macOS (Apple Silicon), enabling local development without requiring Linux/CUDA. Production deployments remain Linux/CUDA only.

 #### Limitations

- macOS support is development-only (no production deployments)
- Single GPU only (parallelism feature disabled -- NCCL not supported)
- Requires Apple Silicon for Metal Performance Shaders
- Some PyTorch operations may fall back to CPU if MPS kernels unavailable

### DIFF
--- a/nix/devShell.nix
+++ b/nix/devShell.nix
@@ -75,6 +75,10 @@
                 # the docs for nix-gl-host say this is a dangerous footgun but.. yolo
                 export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$(${pkgs.nix-gl-host}/bin/nixglhost -p)
               ''
+              + lib.optionalString pkgs.stdenv.isDarwin ''
+                # macOS: Ensure PyTorch can use Metal Performance Shaders
+                export PYTORCH_ENABLE_MPS_FALLBACK=1
+              ''
               + ''
                 echo "Welcome to the Psyche development shell.";
               '';
--- a/nix/devShell.nix
+++ b/nix/devShell.nix
@@ -78,6 +78,9 @@
               + lib.optionalString pkgs.stdenv.isDarwin ''
                 # macOS: Ensure PyTorch can use Metal Performance Shaders
                 export PYTORCH_ENABLE_MPS_FALLBACK=1
+
+                # Set up PyTorch library path for test execution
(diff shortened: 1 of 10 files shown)


arilotter:
> So apparently neither uint16 nor uint32 are supported for PyTorch on Mac M1. Have you found a workaround for this?

Looks like they merged support for unsigned ints last week, though a lot of operations won't work.
https://github.com/pytorch/pytorch/pull/159094

dsocolobsky:
Yeah I saw that, i think I got a workaround for now it seems to be working but not sure if it's training properly (loss goes down tho and it doesn't crash, so at least for development it's useful). Give me a minute.

arilotter:
> We should also make sure that it supports the operations we are performing as in the PR it says not all operations are supported supposedly.

it *should* be totally alright -
the only place we use uint6 & uint32 is in the compression/decompression before immediately recasting as uint8 or Int64, so there should be no operations performed on those datatypes natively :)


I'm going to make a separate PR to upgrade torch to a specific 2.9.0 nightly, including the version check, then ask that this PR be rebased on that.

arilotter:
Oooooooooooookay - pushed a last merge with a TON of changes to another branch to test em
https://github.com/PsycheFoundation/psyche/tree/ari-cursed-merge

:sweat: wanted to clean things up on the cuda device selection side + there was a bunch of weirdness with respect to finetuning that i had to resolve.
Notes:
- Removed the ability to use both MPS and CPU. We already don't support the parallelism feature on non-cuda, so I don't see a reason to support this ATM.
- Refactored devices / optimal devices a bit for cleanliness & typed error handling.
- Added ability to pick specific cuda devices via "cuda:1,3,5" string.
- Moved device parsing to a `FromStr` impl.
- Ran a bunch of stuff through clippy :)



I need to run a couple of the dp/tp/etc changes by some people but other than that this should be good :sweat_smile: 
and then i'll toss that commit on here & merge :grin:

arilotter:
*almost* there. i think this will work: https://github.com/PsycheFoundation/psyche/pull/227

and then i'll merge main into `ari-cursed-merge`, then force-push that here, then merge this to main.

so close!!!!!!

PR #197, opened 2025-08-04
Update setup test script to use permissionless runs

IAvecilla:
This PR updates all the test scripts and the docker container used for the decentralized tests to work with the new way to create a permissionless run.

- Added the `solana-toolbox` in nix dev shell and docker test client image.
- Added the authorizer to the docker test client image to build the idl (had to make it `--impure` as the validator).
- Fix the default wallet usage inside the nix shell because it was trying to use the agenix generated one even when there wasn't a real wallet decrypted there. Now it should use it only if it's there or just use the one in the default location

### DIFF
--- a/docker/test/psyche_solana_validator_entrypoint.sh
+++ b/docker/test/psyche_solana_validator_entrypoint.sh
@@ -12,7 +12,15 @@ solana-test-validator -r &
 sleep 3
 
 pushd /local/solana-authorizer
-anchor deploy --provider.cluster "${RPC}" -- --max-len 500000
+echo -e "\n[+] Creating authorization for everyone to join the run"
+anchor deploy --provider.cluster "${RPC}" --provider.wallet "/.config/solana/id.json" -- --max-len 500000
+sleep 1
+
+anchor idl init \
+    --provider.cluster ${RPC} \
+    --provider.wallet "/.config/solana/id.json" \
+    --filepath /local/solana-authorizer/target/idl/psyche_solana_authorizer.json \
+    PsyAUmhpmiUouWsnJdNGFSX8vZ6rWjXjgDPHsgqPGyw
 popd
 
 pushd /local/solana-coordinator
--- a/docker/test/run_owner_entrypoint.sh
+++ b/docker/test/run_owner_entrypoint.sh
@@ -8,19 +8,7 @@ WALLET_FILE="/root/.config/solana/id.json"
 
 solana airdrop 10 "$(solana-keygen pubkey)"
 
-echo -e "\n[+] - init-idl..."
-pushd /local/solana-authorizer
-anchor idl init \
-    --provider.cluster ${RPC} \
-    --provider.wallet ${WALLET_FILE} \
-    --filepath /local/solana-authorizer/target/idl/psyche_solana_authorizer.json \
-    PsyAUmhpmiUouWsnJdNGFSX8vZ6rWjXjgDPHsgqPGyw
-
-echo -e "\n[+] Creating authorization for everyone to join the run"
-popd
-
 bash /bin/join-authorization-create.sh ${RPC} ${WALLET_FILE} 11111111111111111111111111111111
-
 psyche-solana-client create-run \
     --wallet-private-key-path ${WALLET_FILE} \
     --rpc "${RPC}" \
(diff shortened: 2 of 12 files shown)


scripts/deploy-solana-test.sh
```diff
@@ -20,33 +20,44 @@ echo -e "[+] RUN_ID = $RUN_ID"
...
 
-echo -e "\n[+] starting coordinator deploy"
+echo -e "\n[+] Starting coordinator deploy"
 pushd architectures/decentralized/solana-coordinator
-echo -e "\n[+] syncing keys..."
-anchor keys sync --provider.cluster ${RPC} --provider.wallet $WALLET_FILE
 
-echo -e "\n[+] building..."
+echo -e "\n[+] - syncing keys..."
+anchor keys sync --provider.cluster ${RPC} --provider.wallet ${WALLET_FILE}
```
crypto-vincent: this was there before, but this could be deleted I think ? not sure why this was put in place in the first place. do you have some context ?

IAvecilla: At some point we wanted to make sure that the program id declared in the coordinator matched the key in the `target/deploy` file because we faced a lot of issues with that key changing and the client not being able to communicate with the coordinator. This was before pushing the `key.json` for all the Solana programs so I think now is safe to remove it, I'll do it 👍

IAvecilla:
> note: we should probably also fix: `run_owner_entrypoint.sh` in this same PR but i'm not sure how to test that one

This script was used for the docker setup we use for the decentralized tests. I've updated the script and also the docker images including the `solana_toolbox` and the authorizer so everything should be working now

scripts/join-authorization-create.sh
```diff
@@ -4,6 +4,10 @@ set -o errexit
 set -e
 set -m
 
+set -o errexit
+set -e
+set -m
```
crypto-vincent: nit: duplicate

nix/docker.nix
```diff
@@ -58,6 +59,8 @@ in
       busybox
       cacert
       nixglhostRustPackages."psyche-solana-client-nixglhost"
+      solana_toolbox_cli
+      jq
```
crypto-vincent, line 63: ah yeah thanks for looking into this, it's an implied dependency at the moment since we're using the CLI for everything. Thanks for catching this.

docker/test/run_owner_entrypoint.sh
```diff
@@ -4,24 +4,38 @@ set -o errexit
...
 
 solana airdrop 10 "$(solana-keygen pubkey)"
 
+echo -e "\n[+] - init-idl..."
+pushd /local/solana-authorizer
+anchor idl init \
+    --provider.cluster ${RPC} \
+    --provider.wallet ${WALLET_FILE} \
+    --filepath /local/solana-authorizer/target/idl/psyche_solana_authorizer.json \
+    PsyAUmhpmiUouWsnJdNGFSX8vZ6rWjXjgDPHsgqPGyw
```
crypto-vincent: nit: btw, i don't have context on this but this is only necessary for localnet testing as this is part of the program deployment process (not sure if it belongs here)

IAvecilla: Yes! This script and all the ones inside `docker/test` are used for the decentralized tests so we need to create the permissionless run here also

crypto-vincent: Got it, just wanted to mention that deploying the IDL is not really related to making a permissionless run, it's more related to deploying the authorizer's program itself.

IAvecilla: Okay, I get you now and it totally makes sense. We deploy the authorizer in the `psyche_solana_validator_entrypoint.sh` so that should be the place for the idl deployment.It saves me having the authorizer on the docker client image so it totally makes sense. Thank you!

docker/test/psyche_solana_validator_entrypoint.sh
```diff
@@ -12,7 +12,15 @@ solana-test-validator -r &
 sleep 3
 
 pushd /local/solana-authorizer
-anchor deploy --provider.cluster "${RPC}" -- --max-len 500000
+echo -e "\n[+] Creating authorization for everyone to join the run"
```
crypto-vincent, line 15: micro-nit: here we're deploying the authorizer program, not creating an authorization, but its a comment so it doesn't matter much.

jquesnelle (approved):

PR #208, opened 2025-08-05
pytorch -> 2.9.0 nightly, add flash-attn

arilotter:
Pytorch nightly is required for TorchTitan.
Pytorch newer than 2.8.0 is required for MacOS MPS support.

this also updates all of nixpkgs while we're at it, to get new-enough nccl for the new Pytorch.
closes https://github.com/PsycheFoundation/psyche/issues/203


entropidelic (approved):
lgtm! should we uncomment the `# attn_implementation='flash_attention_2'` in `hf_transformers.py` here too?

PR #215, opened 2025-08-07
Show prompt tasks in frontend

### LINKED ISSUE #198: Show prompt task in run front page
We have already implemented the client task to generate a response given a prompt and send the generated tokens via Witness, see #111 
We still need to:

* Show the generated answer in the run front page.
*Once the prompt reaches the end of sequence, or the task has generated a large number of tokens, we should start again with a new prompt. (Right now we just set the prompt task to finished and stop there)

dsocolobsky:
Modifies the website backend/frontend to show the prompt task results from clients directly on the website. This allows to see actual results produced from the models we're training.

## To test
* Start a run and start training with at least 1 client
* Start the backend and frontend for the website
* You should see a box with the prompt and results in the run dashboard

image

Closes #198


website/backend/src/index.ts
```diff
@@ -333,6 +352,6 @@ async function main() {
 			.send(JSON.stringify(data, replacer))
 	})
 
-	await fastify.listen({ port: 3000 })
+	await fastify.listen({ host: '0.0.0.0', port: 3000 })
```
pefontana, line 355: Can this cause security issues in the server?
I know it's useful for testing, but I think we should not include it in main

website/frontend/src/utils/tokenizer.ts
```diff
@@ -0,0 +1,83 @@
+// Simple LLama2 tokenizer implementation
+let tokenizer: { vocab_reverse: Record<number, string> } | null = null
+
+export async function loadTokenizer() {
+	if (tokenizer) return tokenizer
+
+	try {
+		const response = await fetch('/llama2_tokenizer.json')
```
pefontana: Just to know, can we fetch this from Hugging Face URL directly?
If not, when a run uses another model, the text wont make sense
If it is too complex, we can do it in another PR

shared/client/src/state/stats.rs
```diff
@@ -180,8 +180,9 @@ impl StatsLogger {
             evals
         };
 
-        // See issue https://github.com/PsycheFoundation/psyche/issues/213
-        // let prompt_results = self.get_prompt_results();
+        // TODO see issue https://github.com/PsycheFoundation/psyche/issues/213
```
pefontana: remove the comment
```
// TODO see issue https://github.com/PsycheFoundation/psyche/issues/213
```
I will add the `storage changes` label to the PR

shared/client/src/state/prompt.rs
```diff
@@ -90,13 +135,17 @@ impl PromptTask {
             _ => (),
         }
 
+        // We might want to pass this as config or similar, but it's hardcoded for now.
```
pefontana: I think it is good to leave this hardcoded

scripts/train-solana-test.sh
```diff
@@ -30,6 +30,7 @@ if [[ "$OTLP_METRICS_URL" == "" ]]; then
         --tensor-parallelism ${TP} \
         --micro-batch-size ${BATCH_SIZE} \
         --logs "console" \
+        --prompt-task \
```
pefontana: Remove it, is not always necessary to run the prompt task

shared/client/src/state/stats.rs
```diff
@@ -312,10 +314,24 @@ impl StatsLogger {
                 prompt_task.tokens_to_send.write().unwrap().clear();
             }
         }
-
+        info!(
```
pefontana: This can be a little big and annoying to have in info. We can leave it on trace

shared/coordinator/src/coordinator.rs
```diff
@@ -174,8 +174,9 @@ pub struct WitnessMetadata {
     pub bandwidth_per_sec: f32,
     pub loss: f32,
     pub evals: FixedVec<WitnessEvalResult, 8>,
-    // See issue https://github.com/PsycheFoundation/psyche/issues/213
-    // pub prompt_results: FixedVec<i32, { MAX_TOKENS_TO_SEND }>,
+    // TODO see issue https://github.com/PsycheFoundation/psyche/issues/213
```
pefontana: remove comment

pefontana (changes requested):

website/backend/src/index.ts
dsocolobsky: If I don't recall badly @arilotter told me to add it in a commit. As long as the host has the proper firewall/port config it shouldn't be an issue, but we could pass it via env variable as well.

website/frontend/src/utils/tokenizer.ts
dsocolobsky: True it's a bit hardcoded I'll take a look

dsocolobsky: I think this might take some work, not a lot but it's also not trivial since it's not only changing the tokenizer file but I think the special tokens, whitespace tokens, etc. vary.

I can work on this in this PR though, perhaps I can get it finished before we finish with the rest of the review.

PR #218, opened 2025-08-10
revert torch to 2.7.0

arilotter:
``cargo run --features parallelism --example train -- --tensor-parallelism 8 --distro` will crash with `Segmentation fault: 11`. reverting til we can fix.


jquesnelle:
it's something about the combination of distro backwards + tp (other configs work), need investigation

jquesnelle (approved):

Issue #194, opened 2025-08-01
feat: multinode launcher


jquesnelle:
We can do DP/TP within a node now either as a single process (Rust) or as mixed Rust/Python and pure Python processes (e.g. 1+7). To do fast larger sized runs we need to do multinode parallelism. In both cases, Python and Rust, there is the master Rust process and it either spawns internal threads or new Python processes. We need a simple way to spawn/start the process on other nodes.

**Rust**
```
 master          slave1        slave2
|    R    |     |    R   |     |    R   | 
     ^---------------^--------------^
             dp_comm nccl group
```

**Python**
```
 master          slave1        slave2
|RPPPPPPP|      |PPPPPPPP|     |PPPPPPPP|
 ^^^^^^^^--------^^^^^^^^-------^^^^^^^^
       torch.dist default nccl group
```

Python already has a `sidecar`, namely `psyche.sidecar.__main__` which should more or less work out of the box for each `P` process (need to add global world size is all I think). The question is: how do we get the sidecar launched with the correct arguments on the slave nodes? For Rust we need to make a sidecar executable that does the same thing as the Python sidecar, although would be okay making this TODO if we want to switch to Python trainer fully.

PR #234, opened 2025-08-12
add a standalone sidecar binary & docker integration

arilotter:
This PR adds support for multinode/sidecar training, with a standalone binary for the sidecar nodes.

Within the psyche client docker image, there are two running modes - **main host** and **sidecar**. If the `PSYCHE_MAIN_HOST` is set, the client enters in *sidecar* mode and waits to join the process group of the main host node.

To run the docker client image in sidecar mode, first create an `.sidecar_env` file with the following variables:

```
# .sidecar_env
PSYCHE_MAIN_HOST=<main_host_domain_name>
PSYCHE_WORLD_SIZE=<desired_world_size>
PSYCHE_START_RANK=<starting_rank_of_this_node>
HF_MODEL_REPO=<huggingface_model_repo>
```

where the value for **starting_rank_of_this_node** should be set thinking about the range of ranks this node should spawn. For example, in a multinode run with a world size of 16 and 2 nodes with 8 GPUs each, the sidecar node should use `starting_rank_of_this_node=8` so that it ranges ranks from 8 to 15.

The value for **huggingface_model_repo** should be set with your HuggingFace model repo, for example `NousResearch/Meta-Llama-3.1-8B`. This will pre-download the model inside the docker container on this node.

The recommended way to spawn everything is to first spawn the sidecar in each node and then spawn the main host psyche client.

---

Two important things that change how to run the psyche docker image for multinode run:
* we should run with more shared memory, since the defaults in docker are too small for our use case. Running with `--shm-size=1g` worked fine in my tests but this is something that could be bigger for runs with more nodes/ranks.
* we should share the infiniband device as a volume so that the docker images have high-bandwidth and high-speed communications between them, specially important with the NCCL usage.

As an example, this is how to run the docker sidecar image in the together cluster nodes. Each sidecar node, should run:

```bash
docker run --rm \
    --privileged \
    -v /dev/infiniband:/dev/infiniband \
    --shm-size=1g \
    --gpus all \
    --env-file /tmp/.sidecar_env \
    --network host \
    psyche-solana-client
```

The same **shm-size** and **infiniband** arguments should be passed to the main host node too.

The main host should the same docker command but adjusting the environment file:
```bash
docker run --rm \
	--privileged \
	--shm-size=1g \
	-v /dev/infiniband:/dev/infiniband \
    --gpus all \
    --env-file /tmp/.psyche_env \
    --network host \
    psyche-solana-client
```

where the `.psyche_env` should set the env variable `DATA_PARALLELISM`. If the desired **world size** is for example 16, if `TP=1`, then `DATA_PARALLELISM=16`
 
Closes https://github.com/PsycheFoundation/psyche/issues/194


jquesnelle (approved):

shared/client/src/state/init.rs
```diff
@@ -415,22 +415,16 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static> RunInitConfigAndIO<T
...
                         info!("Loading model...");
 
+                        let serialized_config = source.serialize_config()?;
+
                         let model_task_runner = ModelTaskRunner::new(
                             init_config.eval_tasks,
                             init_config.prompt_task,
                             tokenizer.clone(),
                             init_config.eval_task_max_docs,
-                            // if doing python fsdp we only have one effective dp rank for inference
```
jquesnelle: was this logic removed on purpose? when doing Python FSDP we use the DP parameter but then the DP gets handled on the Python side and there's only one `Trainer` in Rust

PR #250, opened 2025-08-20
evals: add CEval evaluation

### LINKED ISSUE #233: Add an evaluation in another language (chinese)
Add an evaluation in another language, probably chinese
It can be this mmlu evaluation in chinese.
https://github.com/EleutherAI/lm-evaluation-harness/blob/7f04db12d2f8e7a99a0830d99eb78130e1ba2122/lm_eval/tasks/cmmlu/README.md#L11-L12
dataset:  https://huggingface.co/datasets/haonan-li/cmmlu

* Implement the evaluation in Psyche
* Run the evaluation in lm_eval and check that we are getting similar results
* Compare the inputs we are passing to the model vs lm_eval (they should be the same)

dsocolobsky:
Originally in this PR we implemented `cmmlu` however we had some problems with their HF repo and the format of their dataset and now we implement `ceval` which is another evaluation based on multiple-choice answers and log-likelyhood in Chinese, very similar to `cmmlu`.

* `cargo run --release --example evaluate -- --tasks "ceval" --model TinyLlama/TinyLlama-1.1B-Chat-v1.0`:

Closes #233 

The evaluations differ from those in lm_evals when using Deepseek-like models, we believe this to be because of a mismatch in which tokenizer we use, we should investigate this further.
Evals should be very close or the same as those in lm_evals with Llama-like models.


pefontana:
> This happens because we were fetching the branch `revs/convert/parquet` from the HF repo, since we need to handle Parquet format. However they deleted that branch for some reason and it won't work with `main`.
> 
> There's a PR opened from a few days ago to convert to parquet apparently but it has not been merged so let's wait a bit.

ohhh well, so let's wait and merge the PR.
Once they merge it, we should also pin the commit to ensure it doesn't break in the future

One last question, is lm_evals having the same problem?

dsocolobsky:
lm_evals does not have the same problem because they do not use Parquet, they just parse the csv, perhaps we could check if that's too hard to implement here.

shared/eval/examples/evaluate.rs
```diff
@@ -25,15 +25,18 @@ struct Args {
...
+    #[arg(long)]
+    num_fewshot: Option<usize>,
 
     #[arg(long, default_value_t = 42)]
     seed: u64,
 
     #[arg(long, default_value_t = false)]
     quiet: bool,
 
+    #[arg(long)]
```
pefontana, line 40: Is this doing something?

shared/eval/examples/evaluate.rs
```diff
@@ -43,7 +46,16 @@ fn main() -> Result<()> {
     let tasks: Result<Vec<Task>> = args
         .tasks
         .split(",")
-        .map(|x| tasktype_from_name(x).map(|y| Task::new(y, args.num_fewshot, args.seed)))
+        .map(|x| {
+            tasktype_from_name(x).map(|y| {
```
pefontana, line 57: I like it!
It would be nice to add tracing info or a print line if it’s not possible, informing the user about the task name, number of fewshots, and any other info you think would be useful.

shared/eval/examples/evaluate.rs
dsocolobsky: True, I forgot to pass the argument thanks!

shared/eval/examples/evaluate.rs
dsocolobsky: Just added some prints for that!

pefontana (changes requested):
Some questions @dsocolobsky 
1. I dont get why we are using few shots as default, I think lm_eval didnt use any 
```
lm_eval --model hf     --model_args pretrained=deepseek-ai/DeepSeek-V2-Lite   --tasks ceval-valid     --batch_size auto

deep-seek v2 lite
|  Groups   |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|-----------|------:|------|------|--------|---|-----:|---|-----:|
|ceval-valid|      2|none  |      |acc     |↑  |0.5074|±  |0.0132|
|           |       |none  |      |acc_norm|↑  |0.5074|±  |0.0132|
```
2. Have you done some benchmarks, I am gettig really diferent results
```
     Running `target/release/examples/evaluate --tasks ceval --model deepseek-ai/DeepSeek-V2-Lite`
Running tasks with model deepseek-ai/DeepSeek-V2-Lite, seed: 42, DP=1
  - CEval: 4 few-shot examples
CEval: {"acc": 0.6106983655274889, "acc_norm": 0.6106983655274889}
```
```
target/release/examples/evaluate --tasks ceval --model deepseek-ai/DeepSeek-V2-Lite --num-fewshot 0`
Running tasks with model deepseek-ai/DeepSeek-V2-Lite, seed: 42, DP=1
  - CEval: 0 few-shot examples
CEval: {"acc": 0.5891530460624071, "acc_norm": 0.5891530460624071}
```
3. Can we rename the task to `ceval-valid`, so we have the same name as lm_eval
4. I mentioned you to add the acc_norm metric for this task, I was wrong since all answer has length 1 so acc == acc_norm.

dsocolobsky:
actually I've just ran this and it matches with lm_eval so not sure, I actually have changed nothing except for the fewshot default to 0

```
(nix:nix-shell-env) bash-5.3$ cargo run  --release --example evaluate --  --model deepseek-ai/DeepSeek-V2-Lite --tasks ceval-valid
   Compiling psyche-eval v0.1.0 (/mnt/weka/home/lambda/repos/psyche/shared/eval)
    Finished `release` profile [optimized] target(s) in 5.55s
     Running `target/release/examples/evaluate --model deepseek-ai/DeepSeek-V2-Lite --tasks ceval-valid`
Running tasks with model deepseek-ai/DeepSeek-V2-Lite, seed: 42, DP=1
  - CEval-valid: 0 few-shot examples
CEval-valid: {"acc": 0.5891530460624071}
```

shared/eval/examples/evaluate.rs
```diff
@@ -113,14 +139,15 @@ fn run_data_parallel(
...
-    num_fewshot: usize,
     seed: u64,
+    limit: Option<usize>,
     python: bool,
 ) -> Result<()> {
     let task_info: Vec<(String, usize, u64)> = tasks
         .iter()
-        .map(|task| {
-            (format!("{task}"), num_fewshot, seed) // task_name, num_fewshot, seed
+        .enumerate()
```
pefontana: Can we restore this?

pefontana:
okey, so I ran the benchmarks with DeepSeek, and we are getting similar results
```
hf (pretrained=deepseek-ai/DeepSeek-V2-Lite,trust_remote_code=True), gen_kwargs: (None), limit: None, num_fewshot: None, batch_size: auto (64)

|  Groups   |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|-----------|------:|------|------|--------|---|-----:|---|-----:|
|ceval-valid|      2|none  |      |acc     |↑  |0.5966|±  |0.0129|
|           |       |none  |      |acc_norm|↑  |0.5966|±  |0.0129|
```
Note: Since this deepseek model is a little old, we need to set `transformers==4.47.0`

```
~/psyche$ cargo run --release  --example evaluate -- --model deepseek-ai/DeepSeek-V2-Lite --tasks ceval-valid
    Finished `release` profile [optimized] target(s) in 1.95s
     Running `target/release/examples/evaluate --model deepseek-ai/DeepSeek-V2-Lite --tasks ceval-valid`
Running tasks with model deepseek-ai/DeepSeek-V2-Lite, seed: 42, DP=1
  - CEval-valid: 0 few-shot examples
CEval-valid: {"acc": 0.5891530460624071}
```

shared/eval/examples/evaluate.rs
dsocolobsky: good catch done!

pefontana (approved):

PR #251, opened 2025-08-21
evals: mmlu_pro small changes

### LINKED ISSUE #222: GenerateUntil task minor change
GenerateUntil tasks are working OK (see #171), but we need to make a little change to match the lm_eval flow.

We run the GenerateUntil task until we find an answer (the model generates the corresponding regex):
https://github.com/PsycheFoundation/psyche/blob/171/shared/eval/src/harness.rs#L652

In lm_eval, they take a slightly different approach:
They have two different regex patterns - one for stopping token generation and another one to extract the answer the model produced. For example, in mmlu_pro, they generate until they find "Question:" or "</end of sequence>" strings, and then with the generated text they extract the answer.

https://github.com/EleutherAI/lm-evaluation-harness/blob/main/lm_eval/tasks/mmlu_pro/_default_template_yaml#L19-L20
https://github.com/EleutherAI/lm-evaluation-harness/blob/main/lm_eval/models/huggingface.py#L1350

dsocolobsky:
* Changes how we determine the start/end of a new question, by checking for "Question:" instead of finding the answer, this more closely matches lm_eval
* Limits to 1 when a client runs a task for `llmu_pro` since it takes a very long time.

I've ran benchmarks comparing with/without changes and I seem to get the same results.

edit: want to do the benchmarks again just in case since I changed some code when solving merge issues

closes #222


shared/eval/src/tasks/mmlu_pro.rs
```diff
@@ -93,6 +93,15 @@ impl GenerateUntilTask for MMLUPro {
         });
         fewshot_documents
     }
+
+    fn get_stop_tokens(&self) -> Vec<String> {
```
pefontana: This is not tokens, use `get_stop_string` or similar

shared/eval/src/harness.rs
```diff
@@ -674,6 +677,22 @@ impl PreparedTask {
             if generation_complete {
                 cache.write().unwrap().remove(&doc_index);
 
+                // Extract answer from the complete generated text using regex
+                if let Ok(generated_text) = tokenizer.decode(&generated_tokens, false) {
+                    if let Some(captures) = answer_extraction_regex.captures(&generated_text) {
```
pefontana: What happens here if the regex contains more than one match?, We should ensure that we are getting only the last one, or only make the regex with the last part of the generated text

shared/eval/src/tasks/mmlu_pro.rs
```diff
@@ -93,6 +93,15 @@ impl GenerateUntilTask for MMLUPro {
...
+        vec!["Question:".to_string()]
```
pefontana, line 98: I am not 100% sure, but this should not be "\nQuestion:" ?
Check it with lm_evals please

dsocolobsky: In `tasks/mmlu_pro/_default_template_yaml` they have:

```
generation_kwargs:
  until:
    - "Question:"
```

So unless they're adding the newline somewhere else, I think it's without. I can check with/without later in the benchmark perhaps.

shared/eval/src/harness.rs
dsocolobsky: Should be fixed in the last commit, though I want to test in slurm first to check the change was ok

dsocolobsky:
@pefontana ran the final test and got this
```
MMLU Pro: {"acc": 0.24342105263157895}
```

vs in your PR
```
 MMLU Pro: {"acc": 0.24177631578947367}
```

It's a bit different, although I did change how we determine the start/end of question so perhaps that's to be expected? What do you think?

shared/eval/src/tasks/mmlu_pro.rs
pefontana: good!

pefontana:
> @pefontana ran the final test and got this
> 
> ```
> MMLU Pro: {"acc": 0.24342105263157895}
> ```
> 
> vs in your PR
> 
> ```
>  MMLU Pro: {"acc": 0.24177631578947367}
> ```
> 
> It's a bit different, although I did change how we determine the start/end of question so perhaps that's to be expected? What do you think?

Nice, I like it!

pefontana:
@dsocolobsky, just one thing, I just thought we could increase a little the max gen token, in lm_eval they use `max_gen_toks: 2048` I think that is too much, but at least we should allow half of that

https://github.com/PsycheFoundation/psyche/blob/9305f7beba963846cb9eb53f063af0b113c10fa0/shared/eval/src/harness.rs#L17

Everything else seems good

PR #252, opened 2025-08-22
Simulation testing using `Iroh-n0des`

IAvecilla:


IAvecilla:
This got stalled and too complex to actually merge. We're working on examples that could reproduce this exact test, and we might include `iroh-n0des` in the future, but it will be in a separate implementation.

PR #258, opened 2025-08-29
evals: improve fewshot

### LINKED ISSUE #223: Check fewshots are working ok in Log Likelihood tasks
I am not sure that fewshots are working okey in Log Likelihood taks.
Check that the tokens we are passing to the model, match the ones in lm_eval
https://github.com/EleutherAI/lm-evaluation-harness/blob/main/lm_eval/models/huggingface.py#L1310

dsocolobsky:
If you look at the previous code currently in main apparently we do not use `context` for anything, in this PR I attempted to use it in the prompt so that the model gets the actual fewshot context alongside the prompt and options.

We also try to use fewshot from the same category as the question.

I've run some simple benchmarks and it doesn't seem to improve too much if at all, depending on the task, but it seems weird to not use the context for anything. Currently in `main` adding fewshots does not modify the accuracy at all, and now they do.

Closes #223 

image

### DIFF
--- a/shared/eval/src/harness.rs
+++ b/shared/eval/src/harness.rs
@@ -13,7 +13,7 @@ use std::{collections::HashMap, fmt::Display, sync::Arc};
 use tch::{Kind, Tensor};
 use tokenizers::Tokenizer;
 use tokio_util::sync::CancellationToken;
-use tracing::info;
+use tracing::{debug, info};
 const GENERATE_UNTIL_MAX_TOKENS: usize = 600;
 const MAX_CONTEXT_SIZE: usize = 2048;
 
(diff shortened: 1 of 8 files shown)


shared/eval/src/harness.rs
```diff
@@ -180,38 +175,47 @@ impl Task {
...
-                    .iter()
-                    .map(|x| *x as i64)
-                    .collect::<Vec<_>>();
+                // Build individual requests with category-specific fewshot for each document
                 let docs = docs
                     .into_iter()
-                    .map(|x| TokenizedLLHDocument::from_document(x, tokenizer))
+                    .map(|doc| {
+                        // Get fewshot examples for this document's category
+                        let category = doc.category.as_deref().unwrap_or("default");
```
pefontana: Can we move this `let category ...` under the if `let fewshot_prefix = if self.num_fewshot > 0 {`

should make the code easier to read

PR #267, opened 2025-09-08
evaluation: enable DP in evaluate example

### LINKED ISSUE #261: Enable data parallelism in evaluate crate
The `--example evaluate` crate currently runs on a single GPU, which is slow for large datasets. We need to enable data parallelism to use multiple GPUs to speed up the process.

```
# Now
cargo run  --release --example evaluate --  --model pefontana/Tiny-Meta-Llama --tasks mmlu 

# We should be able to do something like
cargo run  --release --features parallelism --example evaluate --  --model pefontana/Tiny-Meta-Llama --tasks mmlu  --data-parallelism 4 
```

It should automatically distribute the data across GPUs and maintain the same accuracy in results, but run faster.

dsocolobsky:
Now one can run
```
cargo run  --release --features parallelism --example evaluate --  --model TinyLlama/TinyLlama-1.1B-Chat-v1.0 --tasks arc_easy --data-parallelism 4
```

I've verified and running with DP=4 takes about 1/4 of the time it takes with DP=1

Closes #261

### DIFF
--- a/shared/eval/examples/evaluate.rs
+++ b/shared/eval/examples/evaluate.rs
@@ -1,9 +1,15 @@
 use anyhow::Result;
 use clap::Parser;
+use psyche_core::RunningAverage;
 use psyche_data_provider::download_model_repo_sync;
 use psyche_eval::{ALL_TASK_NAMES, EvalTaskOptions, Task, tasktype_from_name};
 use psyche_modeling::{auto_model_for_causal_lm_from_pretrained, auto_tokenizer};
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::sync::Arc;
+use std::thread::JoinHandle;
 use tch::{Device, Kind};
+use tokenizers::Tokenizer;
 
 #[derive(Parser, Debug, Clone)]
 struct Args {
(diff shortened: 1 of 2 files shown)


shared/eval/examples/evaluate.rs
```diff
@@ -37,30 +46,135 @@ fn main() -> Result<()> {
...
+                Some(Kind::BFloat16),
+                None,
+                Some(device),
+                None,
+                None,
+            )?;
+
+            for (task_idx, (task_name, num_fewshot, seed)) in task_info.into_iter().enumerate() {
+                let task_type = tasktype_from_name(&task_name)?;
+                let task = Task::new(task_type, num_fewshot, seed + gpu_id as u64);
```
jquesnelle: Don't we want to use the same seed on all ranks but then `skip_and_step_by` handles interleaving? This way the same shuffle happens on all ranks and we just take every `data_parallelism`-th item, offset by `gpu_id`

jquesnelle: @dsocolobsky

dsocolobsky: just seen this, yes it makes sense, running some evals right not to see if this fixes the slight mismatch

dsocolobsky: awesome! with that we get the same `acc` both with DP=1 and DP=4 thanks

PR #272, opened 2025-09-09
NO-MERGE: WIP: fix: indexer memory usage

crypto-vincent:
## Summary

Fix the large resource usage of the indexer by only keeping the data that's necessary

## Details

Specifically we used to store hundreds of thousands of WitnessUpdate data, but really we just need to keep a few thousands at the max and we can sample those over time and sparsify our sample when the list gets too large.


arilotter:
already in main :)

PR #287, opened 2025-09-16
Enable prompt task in docker

pefontana:


shared/client/src/state/prompt_texts.rs
```diff
@@ -12,8 +12,10 @@ struct PromptsJson {
 }
 
 pub fn get_prompt_texts() -> Vec<String> {
-    let json_content = fs::read_to_string("website/frontend/public/prompts/index.json")
-        .expect("Failed to read prompts JSON file");
+    // This prompts folder in created in when we run the client in docker
+    // The file is copied from website/frontend/public/prompts/index.json
+    let json_content =
```
jquesnelle: will this now not work outside of docker? should we just put the file in `prompts/index.json` both in docker and in the repo?

arilotter:
I'm not a huge fan of requiring the binaries to have support text files included alongside them.
I still think an `include_str!` call makes more sense, if we need the exact same text in the eval runner and the frontend anyways.

I also would rather put the files in some common dir, and have the website stuff include the prompt text from the evals rather than the evals include it from the website.

pefontana:
@arilotter @jquesnelle
I addressed the changes you suggested. I tried different approaches and think this is the simplest one.
The prompt texts are inside shared/client/src/state/prompt_texts/index.json, and when we run the front page, we copy them into website/frontend/public/prompts/index.json.
I also pushed a duplicate website/frontend/public/prompts/index.json for the CI to pass, and then added it to .gitignore.

I tested it locally with a Docker client and it worked fine.
Let me know what you think.

pefontana:
@jquesnelle So, @arilotter helped me with Nix, and now we can copy files from the source code into the website.
I just tested it locally, and the prompt task is being displayed on the main page. So I think we’re good to go.

PR #293, opened 2025-09-18
wip: fix: try2 for fixing indexer

crypto-vincent:


nix/devShell.nix
```diff
@@ -43,6 +43,7 @@
 
                 # for some build scripts
                 jq
+                gnused # not installed by default on MacOS!
```
ethernet8023, line 46: this is required for the fixup.sh sed script to work on MacOS, otherwise it'll use the system sed which will break terribly

PR #311, opened 2025-09-25
correctly report download bandwidth

ethernet8023:
new iroh version reports full download amount rather than delta.

### DIFF
--- a/shared/network/src/download_manager.rs
+++ b/shared/network/src/download_manager.rs
@@ -522,9 +522,9 @@ impl<D: Networkable + Send + 'static> DownloadManager<D> {
                     Some(DownloadManagerEvent::Update(DownloadUpdate {
                         blob_ticket: download.blob_ticket.clone(),
                         tag: download.tag,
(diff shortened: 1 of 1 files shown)


shared/network/src/download_manager.rs
```diff
@@ -522,9 +522,9 @@ impl<D: Networkable + Send + 'static> DownloadManager<D> {
                     Some(DownloadManagerEvent::Update(DownloadUpdate {
                         blob_ticket: download.blob_ticket.clone(),
                         tag: download.tag,
-                        downloaded_size_delta: bytes_amount,
+                        downloaded_size_delta: bytes_amount.saturating_sub(download.last_offset),
```
IAvecilla, line 525: I think the `last_offset` is never being re-assigned, since iroh-blobs no longer reports the delta between progress updates. Wouldn’t this always be 0?

ethernet8023: fixed, now we update below :)

jquesnelle (approved):

PR #327, opened 2025-10-08
Add run_id to ServiceInfo so that metrics are annotated

benbarclay:
Injects `run_id` into the ServiceInfo struct as that is consumed by the metrics library when pushing to a metrics host.

Annoyingly, we don't always have access to a run_id though, notably in the centralized server. I've left it blank there, but can easily change that up.

There's likely other impact from this change that I haven't forseen, but from what I've poked around it seems to be safe enough.

### DIFF
--- a/architectures/centralized/server/src/main.rs
+++ b/architectures/centralized/server/src/main.rs
@@ -195,6 +195,7 @@ async fn main() -> Result<()> {
                     namespace: "psyche".to_string(),
                     deployment_environment: std::env::var("DEPLOYMENT_ENV")
                         .unwrap_or("development".to_string()),
+                    run_id: "".to_string(),
                 })
                 .init()?;
             match config {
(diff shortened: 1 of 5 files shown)


telemetry/otel-collector-config.yaml
```diff
@@ -17,6 +17,8 @@ exporters:
 
   prometheus:
     endpoint: '0.0.0.0:8889'
+    resource_to_telemetry_conversion:
+      enabled: true
```
benbarclay, line 21: I also had to add this, as the local telemetry setup was dropping metric annotations by default, which confused me for a while.

architectures/centralized/server/src/main.rs
```diff
@@ -195,6 +195,7 @@ async fn main() -> Result<()> {
                     namespace: "psyche".to_string(),
                     deployment_environment: std::env::var("DEPLOYMENT_ENV")
                         .unwrap_or("development".to_string()),
+                    run_id: "".to_string(),
```
benbarclay: ~I need to brush up on Rust, but I didn't really see a nice way to handle an Optional or something. An empty value here gets dropped when metrics are emitted, but I don't know if it has other side effects elsewhere.~

PR #328, opened 2025-10-08
tools: Run manager

dsocolobsky:
This introduces a new field in `CoordinatorInstanceState` named `client_version` which allows to specify a client version as a string, which will be later used as the tag in the docker container to pull for the client. (Tagged as storage change since it modifies the size of the coordinator slightly).

Introduces a new instruction `UpdateClientVersion` to change the coordinator's client version to something else, in case of an update.

It also introduces a new tool, the `run-manager` which upon running will fetch the coordinator via RPC in order to check the proper client version to download, download the container and run it. The `run-manager` also handles crashes in the inner container, retrying in case of a version mismatch (this allows for re-checking for a new version without exiting the manager).

To test it, one can set the coordinator's client version via the `InitCoordinator` instruction, in local dev one can change the version string in `scripts/deploy-solana-test.sh` here:
```bash
cargo run --release --bin psyche-solana-client -- \
    create-run \
    --wallet-private-key-path ${WALLET_FILE} \
    --rpc ${RPC} \
    --ws-rpc ${WS_RPC} \
    --client-version "latest" \
    --run-id ${RUN_ID} "$@"
```
This will asume the clients run a `nousresearch/psyche-client:latest` container.

Then one can run the `run-manager` as follows:
```bash
cargo run --release -p run-manager -- --env-file config/.env.local --wallet-path some_wallet.json
```

Alternatively, you can specify another version via the `--local someversion` parameter; this is mostly useful for testing purposes, that will use the `psyche-solana-client:someversion` container, if exists.

If you run `--local` without specifiying a tag it will check the coordinator's version and use that, but not pull from DockerHub but rather use from your local registry. Again, this is useful mainly for testing purposes, you can edit `docker.nix` and rebuild with different tags with `just nix build_docker_solana_client`.

## Interesting cases to try
* Happy path: Create a run and run the `run-manager`, it should start the container and start training.
    * If closing via Ctrl+C, no containers should be still running in the background afterwards.
* Mismatched version: Create a run and run the `run-manager` with `--local someotherversion`. It will bypass the initial version check in the run-manager, but once the client connects it will error with mismatched version. The run-manager will keep trying to reconnect.
* Pause and update: Connect a client with the proper version using `--local`, pause the run and change the run's client version via the new instruction `UpdateClientVersion`. The client will detect the update, exit and then the `run-manager` should detect and use the new docker image tag (you might need to build it first via updating the `docker.nix` file here: `tag = local` to something else.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8294,7 +8294,6 @@ dependencies = [
  "psyche-coordinator",
  "psyche-solana-coordinator",
  "solana-client",
- "solana-sdk",
  "tokio",
  "tracing",
  "tracing-subscriber",
--- a/tools/rust-tools/run-manager/Cargo.toml
+++ b/tools/rust-tools/run-manager/Cargo.toml
@@ -21,4 +21,3 @@ clap-markdown.workspace = true
 
 # Solana RPC
 solana-client = "=2.1.4"
-solana-sdk = "=2.1.4"


tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,360 @@
...
+            ));
+        }
+    }
+    Ok(())
+}
+
+/// Get a required environment variable
+fn get_env_var(name: &str) -> Result<String> {
+    std::env::var(name).with_context(|| format!("Missing required environment variable: {}", name))
+}
```
entropidelic: I think that if we just pass the env file, these functions shouldn't be needed

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,360 @@
...
+        info!("Creating container from image: {}", image_name);
+
+        let mut cmd = Command::new("docker");
+        cmd.arg("run")
+            .arg("-d") // detached mode
+            .arg("--network=host")
+            .arg("--shm-size=1g")
+            .arg("--privileged")
+            .arg("--gpus=all")
+            .arg("--device=/dev/infiniband:/dev/infiniband:rwm");
```
entropidelic: is the `rwm` at the end necessary?

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        // Add environment variables
+        for env_var in env_vars {
+            cmd.arg("-e").arg(env_var);
+        }
```
entropidelic: same here, I think this could be removed if using the env file

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,360 @@
...
+        )
+        .init();
+
+    let args = Args::parse();
+
+    // Use default .env file or provided one via parameter
+    let env_file_path = if let Some(ref path) = args.env_file {
+        Some(path.clone())
+    } else {
+        let default_path = PathBuf::from("config/client/.env.local");
```
entropidelic: don't assume any default path, just return if the env file was not provided

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        let mut cmd = Command::new("docker");
+        cmd.arg("run")
+            .arg("-d") // detached mode
+            .arg("--network=host")
+            .arg("--shm-size=1g")
+            .arg("--privileged")
+            .arg("--gpus=all")
+            .arg("--device=/dev/infiniband:/dev/infiniband")
+            .arg("--env")
+            .arg("NVIDIA_DRIVER_CAPABILITIES=all")
```
entropidelic: this env variable is [already set](https://github.com/PsycheFoundation/psyche/blob/10ac987b18fa4d79244372767ff0d043def08b13/nix/docker.nix#L60) in the docker container, so no need to set it here

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,274 @@
...
+        info!("Querying coordinator for Run ID: {}", run_id);
+
+        let instance = self.fetch_coordinator_data(run_id)?;
+        info!(
+            "Fetched CoordinatorInstance from chain: {{ run_id: {}, coordinator_account: {} }}",
+            instance.run_id, instance.coordinator_account
+        );
+
+        // TODO: When version field is added to CoordinatorInstance, use it here:
+        // let docker_tag = format!("nousresearch/psyche-client:{}", instance.version);
```
entropidelic: probably we should add the tag field in this PR?

dsocolobsky: Yes perhaps, does it make sense to add it here at the end or wherever?

```rust
#[derive(Debug, InitSpace)]
#[account]
pub struct CoordinatorInstance {
    pub bump: u8,
    pub main_authority: Pubkey,
    pub join_authority: Pubkey,
    pub coordinator_account: Pubkey,
    #[max_len(SOLANA_MAX_STRING_LEN)]
    pub run_id: String,
}
```

I'll try adding and checking if it breaks anything but probably it will modify the contract size, but we can see about that later.

entropidelic: yeah! I think it makes sense there. We should have in mind how to update this value, I think we may have to add some command to update the docker tag, similar to the other examples we have in `solana-client/src/command`.
cc @arilotter in case you have any thoughts on this

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,393 @@
...
+        let docker_tag = format!("nousresearch/psyche-client:{}", client_version);
+        Ok(docker_tag)
+    }
+}
+
+/// Determine which Docker image to use and pull it if necessary
+async fn prepare_image(args: &Args, docker_mgr: &DockerManager) -> Result<String> {
+    let docker_tag = if let Some(ref local_image) = args.local {
+        info!("Using local Docker image: {}", local_image);
+        local_image.clone()
```
entropidelic: for testing locally I think we should have the same logic here to check for the tag in the coordinator contract

entropidelic: instead of passing the image name as an argument, we could just use the `--local` as a boolean to indicate we should not pull from docker hub

dsocolobsky: This should be done already, you can pass `--local` to fetch the coordinator version and use a local image, or `--local sometag` to force using `sometag` for testing.

architectures/decentralized/solana-client/src/command/update_client_version.rs
```diff
@@ -0,0 +1,45 @@
...
+        &run_id,
+        &coordinator_account,
+        &main_authority,
+        &new_tag,
+    );
+
+    let signature = backend
+        .send_and_retry("Update client version", &[instruction], &[])
+        .await?;
+    // println!("Set pause state to {paused} on run {run_id} with transaction {signature}");
```
entropidelic: we should remove this comment

tools/rust-tools/run-manager/src/main.rs
```diff
@@ -0,0 +1,437 @@
...
+            docker_tag
+        }
+    };
+
+    // Pull image unless --local flag is present
+    if args.local.is_none() {
+        docker_mgr.pull_image(&docker_tag)?;
+    } else {
+        info!("Using local image (skipping pull): {}", docker_tag);
+    }
```
entropidelic: instead of checking this, you can directly return `Ok(docker_tag)` in the `let docker_tag = match  &args.local` branch above

crypto-vincent:
> Hey @crypto-vincent the only thing remaining here would be the fact that this breaks the storage layout, do you have any pointers on what should I change? Thanks!

Hey! that's a tricky one, I think as far as I know there's a few branches that are dormant because making a breaking change in the storage layout would require a complete re-deployment in the current state of things. Maybe lets chat live on this issue since this can't be merged without breaking a whole lot of things (due to storage change)

dsocolobsky:
Right now we're hardcoding the client version in the client's source code, and also in the coordinator. These versions must match for the client to be able to connect.

For each new release we would have to do something like this:
1. Change the version number to `1.0` or whatever in the client and coordinator.
2. Build the docker image and tag it as `:1.0` as the version string should match this.

If let's say the coordinator was updated to `1.2` and the client is outdated for some reason, the run-manager should automatically detect the new version is `1.2` and download the appropiate docker container, inside which the client should have the `1.2` string.

Do you think this approach makes sense?

architectures/decentralized/solana-client/src/app.rs
```diff
@@ -167,13 +169,39 @@ impl App {
...
 
-        let coordinator_account = coordinator_instance_state.coordinator_account;
+        let coordinator_account = coordinator_instance.coordinator_account;
+        let coordinator_account_pubkey = coordinator_instance.coordinator_account;
+        let coordinator_client_version = String::from(
+            &backend
+                .get_coordinator_account(&coordinator_account_pubkey)
+                .await?
+                .state
+                .client_version,
```
ryleung-solana, line 185: Should this client_version not be hardcoded into the coordinator program, and then returned by some instruction instead of being stored in the run PDA, as it will be the same for every run PDA of the coordinator program, and every PDA will need to be updated when the coordinator program gets updated? (see https://solana.stackexchange.com/questions/9361/return-value-from-program-function for how to return data from an instruction from the solana devs). Given that this instruction wouldn't even use any accounts (other than reading the coordinator program's) you could probably even get away with just asking the rpc to simulate the instruction to get the correct result for free (no tx fees) and they shouldn't get madder at this than they would from people asking them to read a PDA, which is also free (https://solana.com/docs/rpc/http/simulatetransaction see the returnData field)

ryleung-solana: As an alternative, you can also just have 1 single PDA associated with the coordinator program reserved to just store the coordinator program's version, although it will need to be updated by someone invoking an instruction on the coordinator program every time a new coordinator program version is deployed. But this will only need to be done once per update, whoever deployed can just remember to do it, and you'll be all set

ethernet8023: > Should this client_version not be hardcoded into the coordinator program, and then returned by some instruction instead of being stored in the run PDA, as it will be the same for every run PDA of the coordinator program, and every PDA will need to be updated when the coordinator program gets updated?

IMO no - it should be okay to update the *client* version without forcing *all existing runs* to stop and download new code.
Maybe we want a *second* version for that, but this version number is "what version of the client code do i need to run to join this", not "is this coordinator ABI-compatible with the previous version", right?

dsocolobsky: Okay yeah that makes sense, we might want to make a quick fix in the clients only and ask the peers to update their clients. In that case the current implementation should do just fine.

ryleung-solana: >  Maybe we want a _second_ version for that, but this version number is "what version of the client code do i need to run to join this", not "is this coordinator ABI-compatible with the previous version", right?

Fair enough though I am still concerned about what happens if the coordinator program is updated... Seems like if we're already implementing a versioning system for the client, we should deal with that as well unless the coordinator is forward-compatible

dsocolobsky: Not sure I understood that last comment

Let's say we have both clients and coordinator at `1.0`.

If we deploy a new version of the coordinator let's say `1.1` then the run manager should see the mismatch and download the right docker image, with tag `:1.1` (we should make sure it exists).

What Ari is proposing is we should be able to update the client only (let's say we fix a critical bug and bump up to `1.2`) without having to redeploy the coordinator if not needed, in which case we just execute the update version command to bump up the version in the coordinator and force the clients to update.

Something to consider is we might want different but compatible versions, let's say for example `1.0.0` and `1.0.2` be compatible with each other and allow joining. But we can check that later.

ryleung-solana: I was just voicing some concern that if we are building a mechanism to ensure you're running the right client version anyway, we might as well also check compatibility with the coordinator program version. But @arilotter 's point taken that the purpose of this is just to make sure all the clients in a single run are running the right version - didn't realize that was the purpose before. If that's the purpose, what's currently here is fine, but I still have the above concerns.

jquesnelle (approved):

Issue #344, opened 2025-10-17
feat: only store the last N checkpoints


arilotter:
right now when we save checkpoints, we will just fill up the disk forever. most trainer systems have the ability to only keep the latest N checkpoints (based on the step number that's in the folder path). we should add an argument to the psyche solana client / the client in general to clean up checkpoints older than the current-Nth checkpoint, at the point in time we'd write a new one.

PR #349, opened 2025-10-21
Add option to limit number of locally saved checkpoints

ryleung-solana:
Add option to the client to specify a number of training checkpoints to keep locally, before deleting any older ones in order to save local disk space. Resolves https://github.com/PsycheFoundation/psyche/issues/344

### DIFF
--- a/website/backend/src/index.ts
+++ b/website/backend/src/index.ts
@@ -297,7 +297,9 @@ async function main() {
 	fastify.get(
 		'/run/:runId/:indexStr',
 		(
-			req: FastifyRequest<{ Params: { runId?: string; indexStr?: string } }>,
+			req: FastifyRequest<{
+				Params: { runId?: string; indexStr?: string }
+			}>,
 			res
 		) => {
 			const isStreamingRequest = req.headers.accept?.includes(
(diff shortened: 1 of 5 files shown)


shared/client/src/state/cooldown.rs
```diff
@@ -189,6 +207,26 @@ impl CooldownStepMetadata {
...
+                        // then we may have more than 1 dir to clean up
+                        while delete_queue_guard.len() > keep_steps as usize {
+                            let delete_step = delete_queue_guard.pop().unwrap().0;
+                            let delete_path =
+                                checkpoint_dir.join(format!("{run_id}-step{delete_step}"));
+                            if let Err(err) = tokio::fs::remove_dir_all(delete_path.clone()).await {
+                                warn!("Error removing {} : {}", delete_path.display(), err);
+                            }
+                        }
+                    }
```
entropidelic: does this mean that local checkpoints won't be deleted if we don't set `keep_steps` explicitly? I think it would make sense to leave some reasonable default value for `keep_steps` so that disk doesn't fill up in case the user forgets to set this argument

ryleung-solana: Good point, I've set some defaults now

PR #350, opened 2025-10-21
website: add checkpoint status check for hf button

IAvecilla:
This PR adds a new component for the Hugging Face button shown on each run page.
Before rendering the button, we make a GET request to the backend with the owner and repo name. The backend then queries Hugging Face, and if the response code is anything other than 200 (OK), the button won’t be displayed.

Additionally, the repo ID must now follow the `"owner/repo"` format (matching Hugging Face’s convention) to perform the status check correctly. For cases such as non-existing, malformed, private, or unauthorized repos, the button will not appear.
This check runs only once when the page loads and will not be triggered again.

### DIFF
--- a/website/frontend/src/components/CheckpointButton.tsx
+++ b/website/frontend/src/components/CheckpointButton.tsx
@@ -12,9 +12,14 @@ export const CheckpointButton = ({
 	const [isValid, setIsValid] = useState<boolean | undefined>(undefined)
 
 	useEffect(() => {
-		const [owner, repo] = checkpoint.repo_id.split('/')
+		const parsedRepo = checkpoint.repo_id.split('/')
 
-		console.log('FETCHING CHECKPOINT STATUS')
+		if (parsedRepo.length !== 2) {
+			setIsValid(false)
+			return
+		}
+		const [owner, repo] = parsedRepo
+
 		fetchCheckpointStatus(owner, repo, checkpoint.revision || undefined)
 			.then((data) => {
 				setIsValid(data.isValid)
@@ -24,7 +29,6 @@ export const CheckpointButton = ({
 			})
 	}, [checkpoint.repo_id, checkpoint.revision])
 
-	// Show nothing while loading
 	if (isValid === undefined) {
 		return null
 	}


dsocolobsky:
I added a log to the fetch function and it seems to be called twice, is that ok?
image

website/frontend/src/routes/runs/$run.$index.tsx
```diff
@@ -85,6 +89,49 @@ function RouteComponent() {
...
+					setCheckpointValidity((prev) => ({
+						...prev,
+						[checkpointKey]: data.isValid,
+					}))
+				})
+				.catch(() => {
+					setCheckpointValidity((prev) => ({ ...prev, [checkpointKey]: false }))
+				})
+			return null
+		}
```
dsocolobsky: I'm not an expert in React but as far as I know I believe these sort of external calls should be done from a `useEffect` and not inside the component, since this chunk of code gets run in each re-render.

I think in this case it's not super important because you check and only do if it's undefined so it _should_ only run once; however I'm not so sure some race condition could happen where it fetches once and then it re-renders before changing the `isValid` property and then it fetches again.

Perhaps you can try if it's not too complex to move to a `useEffect`?

website/backend/src/index.ts
```diff
@@ -365,16 +367,29 @@ async function main() {
...
+		const { owner, repo, revision } = request.query // Changed from request.params
+		const url = `https://huggingface.co/${owner}/${repo}${revision ? `/tree/${revision}` : ''}`
+		try {
+			const response = await fetch(url, { method: 'HEAD' })
+			return { isValid: response.ok, description: response.statusText }
+		} catch (error) {
+			return { isValid: false, description: error.message }
+		}
+	})
+
```
dsocolobsky: I think you can solve the typecheck errors with this

```suggestion
fastify.get<{
		Querystring: { owner: string; repo: string; revision?: string }
	}>('/check-checkpoint', async (request) => {
		const { owner, repo, revision } = request.query
		const url = `https://huggingface.co/${owner}/${repo}${revision ? `/tree/${revision}` : ''}`
		try {
			const response = await fetch(url, { method: 'HEAD' })
			return { isValid: response.ok, description: response.statusText }
		} catch (error) {
			const errorMessage = error instanceof Error ? error.message : 'Unknown error'
			return { isValid: false, description: errorMessage }
		}
	})

```

website/frontend/src/routes/runs/$run.$index.tsx
IAvecilla: Yeah, maybe you're right. I see that we do similar things with useEffect in other parts of the code so I did the same. I ended up moving the `CheckpointButton` to its own file in the components folder and just use it in the run view. Maybe you want to re-review it to make sure that seems okay

jquesnelle (approved):

Issue #288, opened 2025-09-17
Limit run ids to 32 bytes


jquesnelle:
Since we use the run id as the bump seed for the Solana PDAs, we need to limit the seeds to 32 bytes

PR #356, opened 2025-10-28
limit run_id length to 32 bytes

dsocolobsky:
Solves #288 

I've just added a few if checks, before it was failing as well when you passed something >32 bytes however now it's clearer both in code and in the error message what's going on.

**note** the ideal solution would be to limit the field run_id in the Coordinator struct to be 32 bytes and not 64. However, this would imply a struct size change and thus require migrations etc. so for now we can have these checks as a workaround, I did add a comment clarifying that.


jquesnelle:
@pefontana this is now triggering the data layout test

jquesnelle (approved):

PR #359, opened 2025-10-30
fix(bug): distro results tagging

entropidelic:
With the recent blobs tagging changes in the code, both computed and downloaded Distro results were being tagged with the step of that Distro result, which leaded to overwritting of the same tag, leaving the older blob untagged.
In some particular situations, this subtle bug could end up making some client not being able to share its own Distro result and ultimately be dropped from the run. Altough it is a very border case, the fix is necessary for a correct behavior of the blobs garbage collection.

Now the computed Distro results are tagged with `distro-result_{step}`, and the downloaded ones as `downloaded-distro-result-{provider_id}_{step}` so that tags can't step each other. The untagging logic was also changed to reflect this.

Testing the correct garbage collection behavior is difficult, but as a measure to test it, at least in an indirect way, a new env variable was added `BLOBS_GC_INTERVAL_MILLIS` to let us set a very small garbage collection interval in the decentralized integration tests, making errors, if any, appear much more quickly in the tests

### DIFF
--- a/architectures/decentralized/testing/src/docker_setup.rs
+++ b/architectures/decentralized/testing/src/docker_setup.rs
@@ -132,16 +132,18 @@ pub async fn spawn_new_client(docker_client: Arc<Docker>) -> Result<String, Dock
     };
 
     // Get env vars from config file
-    let env_vars = std::fs::read_to_string("../../../config/client/.env.local")
-        .unwrap()
+    let env_vars: Vec<String> = std::fs::read_to_string("../../../config/client/.env.local")
+        .expect("Failed to read env file")
         .lines()
-        .map(|s| s.to_string())
-        .collect::<Vec<String>>();
-    let envs = if has_gpu {
-        [env_vars, vec!["NVIDIA_DRIVER_CAPABILITIES=all".to_string()]].concat()
-    } else {
-        env_vars
-    };
+        .filter_map(|line| {
+            let line = line.trim();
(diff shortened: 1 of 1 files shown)


shared/network/src/lib.rs
```diff
@@ -445,14 +449,47 @@ where
...
+                continue;
+            };
+            let Ok(distro_result_step) = tag_name_distro_result_step.parse::<u32>() else {
+                warn!(
+                    "Distro result step could not be parsed: {tag_name_distro_result_step}. This may lead to a memory leak"
+                );
+                continue;
+            };
+            if distro_result_step < target_distro_result_step {
+                let _ = store.tags().delete(tag_name).await;
```
dsocolobsky: I think this returns a success code

```rust
/// Deletes a tag.
    ///
    /// Returns the number of tags actually removed. Attempting to delete a non-existent tag will *not* fail.
    pub async fn delete(&self, name: impl AsRef<[u8]>) -> super::RequestResult<u64> {
        self.delete_with_opts(DeleteOptions::single(name.as_ref()))
            .await
    }
```

I would check the return value and if it couldn't delete the tag at least log an error.

Also if it couldn't delete the tag, you are adding +1 in the line just below, which is going to cause a mismatch later.

entropidelic: good catch, thanks! solved in e8dd61034fee05a8350dc6a446af84d063e26008

IAvecilla (approved):

PR #360, opened 2025-10-30
test: pause run and rejoin the same client

dsocolobsky:
depends on #363 

Adds a test that pauses a run, leaves the run without any clients and then the same client rejoins.

`just decentralized-integration-test test_pause_and_resume_run`


dsocolobsky:
Main logic should be mostly ready for review though I wanted to clean up code and reduce some duplicated code I added.

architectures/decentralized/testing/src/docker_setup.rs
```diff
@@ -167,6 +169,106 @@ pub async fn spawn_new_client(docker_client: Arc<Docker>) -> Result<String, Dock
...
+         solana config set --url \"${RPC}\" && \
+         solana airdrop 10 \"$(solana-keygen pubkey)\" && \
+         psyche-solana-client train \
+             --wallet-private-key-path \"/root/.config/solana/id.json\" \
+             --rpc \"${RPC}\" \
+             --ws-rpc \"${WS_RPC}\" \
+             --run-id \"${RUN_ID}\" \
+             --logs \"json\""
+        .to_string();
+    let entrypoint = vec!["/bin/sh", "-c", &entrypoint_cmd];
```
entropidelic: maybe we should just move this to a separate bash script and pass it as entrypoint

architectures/decentralized/testing/src/docker_setup.rs
```diff
@@ -333,3 +435,96 @@ pub async fn kill_all_clients(docker: &Docker, signal: &str) {
...
+        .map_err(|e| format!("Failed to create file {}: {}", output_path, e))?;
+    file.write_all(keypair_json.as_bytes())
+        .map_err(|e| format!("Failed to write keypair: {}", e))?;
+
+    println!(
+        "Extracted keypair from {} to {}",
+        container_name, output_path
+    );
+    Ok(())
+}
```
entropidelic: instead of extracting the key from a container, can't we just use a pre-defined one and use it when spawning the docker instance?

architectures/decentralized/testing/src/docker_setup.rs
```diff
@@ -35,6 +36,34 @@ pub const CLIENT_CONTAINER_PREFIX: &str = "test-psyche-test-client";
...
+        return Err(format!(
+            "solana-keygen failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        ));
+    }
+
+    println!("Generated keypair at {}", output_path);
+    Ok(())
+}
+
```
entropidelic: we can just create a key and store it in the testing folder to use it the tests

architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/instance_state.rs
```diff
@@ -96,6 +96,21 @@ impl CoordinatorInstanceState {
...
+                // Log pending client IDs for debugging
+                if active_clients_ids.len() > 0 {
+                    msg!("Pending client IDs:");
+                    for client in self.clients_state.clients.iter() {
+                        if client.active == self.clients_state.next_active {
+                            msg!("  {:?}", client.id);
+                        }
+                    }
+                }
+
```
entropidelic: lets remove these if not necessary

architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/instance_state.rs
```diff
@@ -170,6 +185,22 @@ impl CoordinatorInstanceState {
...
+                    psyche_coordinator::model::Checkpoint::P2P(_) => "P2P",
+                    psyche_coordinator::model::Checkpoint::Ephemeral => {
+                        "Ephemeral"
+                    },
+                    psyche_coordinator::model::Checkpoint::Dummy(_) => "Dummy",
+                };
+                msg!("Checkpoint type: {}", checkpoint_str);
+            },
+        }
+
```
entropidelic: same here

architectures/decentralized/testing/src/docker_setup.rs
```diff
@@ -145,55 +214,196 @@ pub async fn spawn_new_client(docker_client: Arc<Docker>) -> Result<String, Dock
...
+    spawn_container_with_keypair_and_config(
+        docker_client,
+        container_name,
+        host_keypair_path,
+        entrypoint,
+        Some(config_path), // Run owner needs config file
+        additional_env_vars,
+    )
+    .await
```
entropidelic: having the hardcoded key, can we just use it always without having to specify it? in that case we could just remove this function

shared/coordinator/src/coordinator.rs
```diff
@@ -644,6 +644,8 @@ impl<T: NodeIdentity> Coordinator<T> {
             if self.active() {
                 self.pending_pause = true.into();
             } else {
+                println!("withdraw_all 1");
+                self.withdraw_all()?;
```
entropidelic: check these, I think the fix is not correct here

shared/coordinator/src/coordinator.rs
```diff
@@ -1027,9 +1029,10 @@ impl<T: NodeIdentity> Coordinator<T> {
             }
 
             if self.pending_pause.is_true() {
+                println!("withdraw_all 2");
+                self.withdraw_all()?;
```
entropidelic: same here

architectures/decentralized/testing/src/docker_setup.rs
```diff
@@ -65,6 +94,46 @@ pub async fn e2e_testing_setup(
...
+            }),
+        )
+        .await;
+    println!("Removed docker-compose run owner container");
+
+    spawn_ctrl_c_task();
+
+    DockerTestCleanup {}
+}
+
```
entropidelic: let's rethink if this is still necessary if using a hardcoded solana key

shared/coordinator/src/coordinator.rs
```diff
@@ -644,6 +644,7 @@ impl<T: NodeIdentity> Coordinator<T> {
             if self.active() {
                 self.pending_pause = true.into();
             } else {
+                self.withdraw_all()?;
```
jquesnelle: what is the effect of this? we'll treat all nodes as withdrawn when we pause? we may need to special case this somehow so rewards still get handed out

entropidelic: this is actually the fix from #363, we just added the fix here to make sure the test worked

entropidelic: maybe we should add another method to eject all clients without penalizing them?

PR #361, opened 2025-10-30
refactor: Improvements on download manager module

IAvecilla:
This PR aims to simplify the `download manager` module. The main organizational and logical changes are:

* The blob store now lives inside the download manager instead of the networking layer. This prevents the networking side from having to read downloaded blobs just to notify the download manager of completion.
* The download manager now has a main function called `start_download`, which spawns a new Tokio task that tracks updates from an Iroh download. It sends these updates to the networking layer to update metrics and download states. When a download completes, the download manager reads the data from the blob store and sends a completion update to the networking layer.
* All blob store–related operations such as uploading blobs for peers to download, tagging currently downloading blobs, and removing tags from older blobs are now handled by the download manager. The networking side simply calls these functions when needed.


shared/network/src/download_manager.rs
```diff
@@ -346,301 +323,253 @@ impl<D: Networkable> Debug for DownloadManagerEvent<D> {
...
+            let _ = store.tags().set(tag.clone(), blob_ticket.hash()).await;
+            info!(
+                "Starting download for blob: {}",
+                download.blob_ticket.hash()
+            );
+            let progress = progress.stream().await;
+            match progress {
+                Ok(mut progress) => {
+                    while let Some(val) = progress.next().await {
+                        match val {
```
entropidelic: since we are just checking for one variant, can we just use `matches!()`? I think it will make this a little simpler

shared/network/src/download_manager.rs
```diff
@@ -346,301 +323,253 @@ impl<D: Networkable> Debug for DownloadManagerEvent<D> {
...
+                            DownloadProgressItem::PartComplete { request: _request } => {
+                                let event = Self::process_completed(
+                                    blobs,
+                                    download.blob_ticket.clone(),
+                                    tag.clone(),
+                                    download_type.clone(),
+                                )
+                                .await;
+                                if let Some(event) = event {
+                                    let _ = event_sender.send(event);
```
entropidelic: can this error? maybe we can just log the error if it happens and continue

shared/network/src/download_manager.rs
```diff
@@ -346,301 +323,253 @@ impl<D: Networkable> Debug for DownloadManagerEvent<D> {
...
+                                }
+                                break;
+                            }
+                            other => {
+                                if let Some(event) =
+                                    Self::handle_download_progress(other, &mut download)
+                                {
+                                    let _ = event_sender.send(event);
```
entropidelic: same here

shared/network/src/download_manager.rs
IAvecilla: Yes! good idea

PR #362, opened 2025-10-30
Fix checkpoint uploads when pausing clients or run out of cooldown time

pefontana:
When a client pauses the run, it was closing before finishing the Hugging Face upload, causing incomplete
checkpoints. Also, if the upload didn't finish within the cooldown time, it would fail entirely.

This PR changes the checkpoint flow to spawn HF uploads in a separate background thread. The client now tracks these
pending uploads and waits for them to complete before closing when paused. If the upload doesn't finish during
cooldown, it continues in the background during the next epoch instead of failing.

Main changes:
- Refactored cooldown.rs to spawn uploads in separate async tasks 
- Refactored `fn while run.doing_checkpoint()` to track the upload thread progress
- Added pending_upload_handles to track background uploads in steps.rs
- Removed the unused CoordinatorEpochState.checkpointed field
- Since now the checkpoint N can be uploaded in the N+1 epoch, I modify the client arg `keep_steps` to be 1 or more when the client is a checkpointer

This ensures checkpoints are fully uploaded and available even when clients pause during cooldown or when uploads take
longer than the cooldown period.

### DIFF
--- a/shared/coordinator/src/coordinator.rs
+++ b/shared/coordinator/src/coordinator.rs
@@ -635,6 +635,7 @@ impl<T: NodeIdentity> Coordinator<T> {
             if self.active() {
                 self.pending_pause = true.into();
             } else {
+                self.withdraw_all()?;
                 self.change_state(unix_timestamp, RunState::Paused);
                 self.epoch_state.cold_start_epoch = true.into();
             }
@@ -792,11 +793,14 @@ impl<T: NodeIdentity> Coordinator<T> {
     }
 
     pub fn active(&self) -> bool {
-        !self.halted()
-            && !matches!(
-                self.run_state,
-                RunState::WaitingForMembers | RunState::Warmup
-            )
+        !matches!(
+            self.run_state,
+            RunState::WaitingForMembers
+                | RunState::Warmup
+                | RunState::Uninitialized
+                | RunState::Finished
+                | RunState::Paused
+        )
     }
 
     pub fn halted(&self) -> bool {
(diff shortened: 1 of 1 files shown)


jquesnelle:
@pefontana weird this isn't a storage change, didn't we remove the `checkpointed` variable?

shared/coordinator/src/coordinator.rs
```diff
@@ -407,7 +404,6 @@ impl<T: NodeIdentity> Default for CoordinatorEpochState<T> {
             rounds: Default::default(),
             rounds_head: Default::default(),
             first_round: true.into(),
-            checkpointed: Default::default(),
```
jquesnelle, line 410: this is a storage change right?

pefontana: I though the same, but since the test was not failing I assumed that no

jquesnelle: @crypto-vincent

dsocolobsky: I might be wrong but I have a feeling that since we've just removed an `u8` from the struct it might be reading either the previous or following field (so an `u8` as well) into that field? So the deserialize is not crashing but we might have corrupted data.

dsocolobsky: What I've seen people do is instead of removing a field they change the name to `_unused` or `_padding` or whatever and leave the same size as padding. Should not break things.

But since we're adding fields anyways I guess we might as well delete it.

jquesnelle (approved):

Issue #357, opened 2025-10-29
Client fails to get model via P2P when rejoining paused run


pefontana:
When a run is paused and then resumed, if a client rejoins the run using the same Solana address, it attempts to fetch the model via P2P first, which fails. The client only succeeds in getting the model via the Hub on the second rejoin attempt.

When rejoining a paused run with the same Solana address, the client should directly get the model from the Hub to avoid the p2p failure.

PR #363, opened 2025-10-31
Fix pause bug

pefontana:
Right now, we have a bug when a client rejoins a run that was paused. See: #357
These changes fix it.
Now, when the run is Paused  it withdraw all the clients. Then, when the run is resumed and the clients join, they get the model via Hub

Also, 
* Refactor `fn Coordinator::paused` to make it easier to read
* Fix a bug in `fn get_cold_start_warmup_bounds` when `cold_start_warmup_steps == 0`


shared/coordinator/src/coordinator.rs
```diff
@@ -857,16 +861,16 @@ impl<T: NodeIdentity> Coordinator<T> {
...
-                self.epoch_state.start_step,
-                self.epoch_state.start_step
-                    + match &self.model {
-                        Model::LLM(llm) => llm.cold_start_warmup_steps,
-                    },
-            )),
-            false => None,
+        let cold_start_warmup_steps = match &self.model {
+            Model::LLM(llm) => llm.cold_start_warmup_steps,
+        };
```
entropidelic: pico nit :)
```suggestion
        let Model::LLM(llm) = &self.model;
        let cold_start_warmup_steps = llm.cold_start_warmup_steps;
```

entropidelic (approved):

shared/coordinator/src/coordinator.rs
pefontana: Nice one!
Didnt know that one

PR #365, opened 2025-11-05
feat: Epoch rewards shared between clients instead of "per_client"

crypto-vincent:
## Summary

As discussed, one change that we want to make to the epoch's rewards be a fixed a mount per-epoch instead of depending on the number of finishing clients.

## Details

This PR simply rename the field to clarify the implementation logic, and simply divide the reward rate per number of finishing client at the end of the epoch

### DIFF
--- a/.github/actions/wait-for-garnix/action.yaml
+++ b/.github/actions/wait-for-garnix/action.yaml
@@ -1,0 +1,50 @@
+name: 'Wait for Garnix CI'
+description: 'Wait for Garnix CI checks to complete'
+inputs:
+  timeout-minutes:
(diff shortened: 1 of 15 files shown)


jquesnelle:
@crypto-vincent is there anything with the audit you'd feel nervous about with this change?

crypto-vincent:
> @crypto-vincent is there anything with the audit you'd feel nervous about with this change?

There's no logic change in the treasurer itself which is the only part that's currently been audited, so I would say we are safe on that front. (The coordinator hasnt been audited yet)

This is the only change in the treasurer folder (just a variable rename for clarity):
Screenshot 2025-11-12 at 20 11 47

PR #378, opened 2025-11-14
ci: add python decentralized tests to run in the CI

### LINKED ISSUE #415: tests: add test with python feature and sidecar usage
Add a test to the CI that runs only when using the `python` feature using sidecars and p2p model sharing to check the rank broadcast works as expected. For this we also need to set up our self hosted runner with more than one GPU and enough memory to build all the Python related packages.

IAvecilla:
This closes #415

### DIFF
--- a/architectures/inference-only/inference-node/Cargo.toml
+++ b/architectures/inference-only/inference-node/Cargo.toml
@@ -36,6 +36,7 @@ iroh.workspace = true
 iroh-blobs.workspace = true
 iroh-gossip.workspace = true
 
+reqwest = { version = "0.12", features = ["json"] }
 serde_json.workspace = true
 serde.workspace = true
 uuid = { version = "1", features = ["v4"] }
--- a/architectures/inference-only/inference-node/src/bin/gateway-node.rs
+++ b/architectures/inference-only/inference-node/src/bin/gateway-node.rs
@@ -14,9 +14,10 @@ use axum::{
     extract::State,
     http::StatusCode,
     response::{IntoResponse, Response},
-    routing::post,
+    routing::{get, post},
 };
 use clap::Parser;
+use iroh::EndpointAddr;
 use psyche_inference::{
     INFERENCE_ALPN, InferenceGossipMessage, InferenceMessage, InferenceRequest, InferenceResponse,
 };
@@ -68,6 +69,7 @@ struct GatewayState {
     pending_requests: RwLock<HashMap<String, mpsc::Sender<InferenceResponse>>>,
     network_tx: mpsc::Sender<(EndpointId, InferenceMessage)>,
     gossip_tx: mpsc::Sender<InferenceGossipMessage>,
+    endpoint_addr: EndpointAddr,
 }
 
 #[derive(serde::Deserialize, serde::Serialize, Clone, Debug)]
@@ -271,6 +273,15 @@ async fn handle_load_model(
     ))
 }
 
+#[axum::debug_handler]
+async fn handle_bootstrap(State(state): State<Arc<GatewayState>>) -> Json<EndpointAddr> {
+    info!(
+        "Bootstrap request: returning endpoint addr {}",
+        state.endpoint_addr.id.fmt_short()
+    );
+    Json(state.endpoint_addr.clone())
+}
+
 #[derive(Debug)]
 enum AppError {
     NoNodesAvailable,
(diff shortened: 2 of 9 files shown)


.github/workflows/solana-decentralized-integration-test-run.yml
```diff
@@ -0,0 +1,82 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Get Validator Image from cache
+        id: cache-validator
+        uses: actions/cache/restore@v4
+        with:
+          path: validator-image.tar.gz
+          key: validator-image-${{ runner.os }}-${{ hashFiles('architectures/decentralized/solana-coordinator/**/*.rs', 'architectures/decentralized/solana-coordinator/**/*.toml', 'architectures/decentralized/solana-coordinator/Cargo.lock', 'architectures/decentralized/solana-authorizer/**/*.rs', 'architectures/decentralized/solana-authorizer/**/*.toml', 'architectures/decentralized/solana-authorizer/Cargo.lock', 'docker/test/psyche_solana_validator_entrypoint.sh', 'nix/docker.nix', 'flake.lock') }}
```
dsocolobsky: Perhaps I would add `shared/coordinator/src/coordinator.rs` like we did in https://github.com/PsycheFoundation/psyche/pull/426

dsocolobsky: in fact I think we might need to or the key is going to be different

.github/workflows/solana-integration-test-run.yml
```diff
@@ -36,3 +41,65 @@ jobs:
...
+          rm -f validator-image.tar.gz
+          echo "Disk usage after removing validator tar"
+          df -h
+
+      - name: Download Solana Test Client Python Image
+        run: |
+          echo "Disk space before client build"
+          df -h
+
+          sleep 500
```
pefontana, line 91: ?

.github/workflows/solana-integration-test-run.yml
```diff
@@ -36,3 +41,65 @@ jobs:
     with:
       test-name: ${{ matrix.test-name }}
     secrets: inherit
+
+  decentralized-integration-python-test:
+    runs-on: self-hosted
+    needs: build-validator
```
pefontana, line 49: Can we add a tomeout here?
`timeout-minutes: 60`

.github/workflows/solana-integration-test-run.yml
```diff
@@ -36,3 +41,65 @@ jobs:
...
+          $OUT_PATH | docker load
+
+          echo "Disk space after client build"
+          df -h
+
+      - name: Run decentralized integration test
+        env:
+          HF_TOKEN: ${{ secrets.HF_TOKEN }}
+        run: |
+          nix develop .#dev-python --command bash -c "cargo test --release --features python,parallelism -p psyche-decentralized-testing --test integration_tests -- --nocapture test_run_with_python"
```
pefontana, line 107: It will be ok to make a cleanup after the run. Or will slow the testing?
```
      - name: Cleanup
        if: always()
        run: |
          docker system prune -af
          rm -rf validator-image.tar.gz 2>/dev/null || true
```

.github/workflows/solana-integration-test-run.yml
IAvecilla: This is just a quick way to wait for the garnix action that builds the client image to finish, before merge we should install gh cli tool to use the `wait-for-garnix` action we already have to actually wait for that

.github/workflows/solana-integration-test-run.yml
IAvecilla: We already planned a service to run every 24 hours in the server that will do `docker system prune` and `nix-collect-garbage` and keep the disk clean I think that will cover this

.github/workflows/solana-integration-test-run.yml
IAvecilla: Sure, good idea

docker/test/client_test_entrypoint.sh
```diff
@@ -16,8 +16,9 @@ if [ "${PYTHON_ENABLED}" = "true" ]; then
         --rpc "${RPC}" \
         --ws-rpc "${WS_RPC}" \
         --run-id "${RUN_ID}" \
-        --data-parallelism 8 \
+        --data-parallelism 2 \
         --sidecar-port "${SIDECAR_PORT}" \
+        --iroh-relay "n0" \
```
pefontana: is ok to use this relay?

IAvecilla: No, that was wrong. I already updated it!

PR #383, opened 2025-11-21
Correct OLTP typo

benbarclay:
This change replaces OLTP with the correct abbreviation OTLP. 
It also provides a fallback for when the OLTP env vars are still set.

This should resolve https://github.com/PsycheFoundation/psyche/issues/192


benbarclay:
I've only kept both as an intermediate step while the change is rolled out. If we want to make it a hard breaking change, then that's pretty trivial to do so as well.

PR #402, opened 2025-12-07
vLLM Python integration to enable real-time weight updates

samherring99:
This PR adds several files to the `python/python/psyche/vllm` subpath to enable vLLM to run in inference nodes for Psyche RL runs. The high level plan is to integrate the torchtitan GRPO logic in a new trainer to coordinate weight updates with Psyche inference nodes,

Main files added and their purpose:

### `add_request(prompt: str, sampling_params_dict: Dict) -> str`

  - Adds inference request to vLLM queue
  - Converts dict to SamplingParams object
  - Returns auto-generated request ID

### `step() -> List[RequestOutput]`

  - Executes one inference step
  - Returns list of completed/partial request outputs
  - Returns empty list if no unfinished requests

### `load_model(eep_scale_up=False)`

  - Calls parent's load_model() first
  - Calls model.share_memory() to enable PyTorch shared memory
  - Gets state_dict and calls share_memory_() on each tensor
  - Stores reference as self.psyche_shared_state_dict
  - Spawns distributed updater process

### ` _spawn_distributed_updater(state_dict)`

  - Reads config from environment:
  - Spawns daemon process using spawn context
  - Passes shared memory state_dict to subprocess

### `weight_updater_process(state_dict, process_group_config, model_config)`

  - Entry point for the spawned updater subprocess
  - Runs as daemon process in background
  - Calls `dist.init_process_group()` to join training process group
  - Waits for metadata broadcast from rank 0
  - Receives parameter tensor broadcast and applies with `state_dict[param_name].data.copy_(param_tensor)`

### `broadcast_parameter(param_name, param_tensor, src_rank=0, group=None)`

- Broadcasts a single parameter to all ranks in process group
- Broadcasts metadata tensor

### `broadcast_state_dict(state_dict, src_rank=0, group=None)`

  - Broadcasts entire model state dict parameter-by-parameter
  - Calls broadcast_parameter() for each tensor in dict

## `test_weight_updates.py` - end-to-end test demonstrating the full distributed weight update

To verify this works and set up yourself, run the following:
```
nix develop .#dev-python
python -m venv .venv
source .venv/bin/activate

uv pip install vllm
uv pip install -e python

python -m test_weight_updates
```

**If `_psyche_ext.so` is missing**
```
ext_path=$(find /nix/store -name "_psyche_ext.so" -path "*/psyche/_psyche_ext.so" 2>/dev/null | head -1)
cp "$ext_path" python/python/psyche/_psyche_ext.so
``` 

The next step of this integration is to build a Rust bridge to broadcast weight udpates from Psyche through PyO3.

### DIFF
--- a/python/python/psyche/vllm/vllm_patch.py
+++ b/python/python/psyche/vllm/vllm_patch.py
@@ -48,18 +48,14 @@ def apply_vllm_patches():
 
             def _spawn_distributed_updater(self, state_dict):
                 from psyche.vllm.distributed_updater import weight_updater_process
-                import os
 
-                if "PSYCHE_WORLD_SIZE" not in os.environ:
-                    return
-
+                # Local process group: broadcaster (rank 0) + updater (rank 1)
+                # Use gloo - reliable, simple, fast enough for localhost broadcasts
                 process_group_config = {
-                    "backend": os.environ.get("PSYCHE_UPDATER_BACKEND", "nccl"),
-                    "init_method": os.environ.get(
-                        "PSYCHE_UPDATER_INIT_METHOD", "env://"
-                    ),
-                    "world_size": int(os.environ["PSYCHE_WORLD_SIZE"]),
-                    "rank": int(os.environ["PSYCHE_RANK"]),
+                    "backend": "gloo",
+                    "init_method": "tcp://127.0.0.1:29500",
+                    "world_size": 2,  # Always 2: local broadcaster + this updater
+                    "rank": 1,  # Updater is always rank 1
                 }
 
                 ctx = mp.get_context("spawn")
--- a/python/python/psyche/vllm/__init__.py
+++ b/python/python/psyche/vllm/__init__.py
@@ -1,3 +1,7 @@
+"""
+vLLM integration for Psyche.
+"""
+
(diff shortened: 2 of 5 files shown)


python/python/psyche/vllm/vllm_patch.py
```diff
@@ -0,0 +1,77 @@
...
+
+                self.psyche_shared_state_dict = state_dict
+
+                self._spawn_distributed_updater(state_dict)
+
+            def _spawn_distributed_updater(self, state_dict):
+                from psyche.vllm.distributed_updater import weight_updater_process
+
+                # Local process group: broadcaster (rank 0) + updater (rank 1)
+                # Use gloo - reliable, simple, fast enough for localhost broadcasts
```
dmahan93: CPU is nowhere near fast enough for real-time weight updates, as a step towards it sure this is fine for debugging since we'll be using distro's optimizer stuff or reverting back to having TT/vLLM live on the same interconnect

samherring99: Ok thank you I wondered about this part, will look into getting nccl backend to work here.

samherring99:
Closing this PR in favor of moving to Psyche integration faster, so as to not be blocked on `torch.distributed` integrations.

PR #403, opened 2025-12-07
Add update client version treasurer

jquesnelle:


architectures/decentralized/solana-client/src/command/update_config.rs
```diff
@@ -151,21 +151,30 @@ pub async fn command_update_config_execute(
...
-            config,
-            model,
-            progress,
-        )
+        vec![
+            instructions::coordinator_update_client_version(
+                &run_id,
+                &coordinator_account,
+                &main_authority,
+                "test-client-version",
```
crypto-vincent: nit: might want to leave a todo to update this programatically or when doing some client changes ?

jquesnelle: actually hold on yeah this might be wrong

PR #405, opened 2025-12-09
update reference storage layout for coordinator v1

jquesnelle:


architectures/decentralized/solana-tooling/tests/fixtures/coordinator-account-v1.so
crypto-vincent, line 1: micro-question: we could just replace the old layout? any thought as to why you wanted to keep both ?

PR #420, opened 2025-12-09
docs: Update book with latest changes

IAvecilla:
- Update all the docs in the Psyche book to match with the latest changes
- Test and update the commands and scripts used for development
- Create new scripts to make it easier to test both in localnet and devnet
- Add a new Justfile module for development under the `just dev` module

### DIFF
--- a/.github/workflows/solana-integration-test-base.yml
+++ b/.github/workflows/solana-integration-test-base.yml
@@ -41,8 +41,7 @@ jobs:
 
       - name: Install just
         run: |
-          curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to /usr/local/bin
-          sudo chmod +x /usr/local/bin/just
+          nix profile install nixpkgs#just
 
       # Step 1: Get Validator Image from cache
       - name: Get Validator Image from cache
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,6 +4,7 @@ members = [
   "architectures/decentralized/solana-client",
   "architectures/decentralized/testing",
   "architectures/decentralized/solana-tooling",
+  "architectures/inference-only/*",
   "python/",
   "python/extension-impl/",
   "shared/*",
--- a/architectures/inference-only/inference-node/Cargo.toml
+++ b/architectures/inference-only/inference-node/Cargo.toml
@@ -1,0 +1,34 @@
+[package]
+name = "psyche-inference-node"
+version.workspace = true
+edition = "2021"
+
+[[bin]]
+name = "psyche-inference-node"
+path = "src/main.rs"
+
+[[bin]]
+name = "test-network"
+path = "src/bin/test-network.rs"
+
+[dependencies]
+psyche-inference.workspace = true
+psyche-network.workspace = true
+psyche-core.workspace = true
+psyche-metrics.workspace = true
+
+tokio.workspace = true
+tokio-util.workspace = true
+
+anyhow.workspace = true
+
+tracing.workspace = true
+tracing-subscriber = { version = "0.3", features = ["env-filter"] }
+
+clap = { version = "4", features = ["derive"] }
+
+iroh.workspace = true
+iroh-blobs.workspace = true
+iroh-gossip.workspace = true
+
+serde_json.workspace = true
--- a/architectures/inference-only/inference-node/src/bin/test-network.rs
+++ b/architectures/inference-only/inference-node/src/bin/test-network.rs
@@ -1,0 +1,196 @@
+//! Test binary for P2P network without vLLM
+//!
+//! Usage:
+//!   Terminal 1: cargo run --bin test-network -- --node-id 1
+//!   Terminal 2: cargo run --bin test-network -- --node-id 2
+//!
+//! They should discover each other and see availability announcements.
+
+use anyhow::{Context, Result};
+use clap::Parser;
+use iroh::EndpointAddr;
+use psyche_inference::InferenceGossipMessage;
+use psyche_metrics::ClientMetrics;
+use psyche_network::{DiscoveryMode, NetworkConnection, NetworkEvent, RelayKind, allowlist};
+use std::{fs, path::PathBuf, sync::Arc, time::Duration};
+use tokio::time::sleep;
+use tokio_util::sync::CancellationToken;
+use tracing::{debug, error, info};
+
+#[derive(Parser, Debug)]
+struct Args {
+    #[arg(long)]
+    node_id: String,
+
+    #[arg(long, default_value = "local")]
+    discovery_mode: String,
+
+    #[arg(long, default_value = "disabled")]
+    relay_kind: String,
+
+    #[arg(long)]
+    bootstrap_peer_file: Option<PathBuf>,
+
+    #[arg(long)]
+    write_endpoint_file: Option<PathBuf>,
+}
+
+#[tokio::main]
+async fn main() -> Result<()> {
+    tracing_subscriber::fmt()
+        .with_env_filter(
+            tracing_subscriber::EnvFilter::try_from_default_env()
+                .unwrap_or_else(|_| tracing_subscriber::EnvFilter::new("info")),
+        )
+        .init();
+
+    let args = Args::parse();
+
+    info!("Starting test node: {}", args.node_id);
+
+    let discovery_mode: DiscoveryMode = args
+        .discovery_mode
+        .parse()
+        .map_err(|e| anyhow::anyhow!("Invalid discovery mode: {}", e))?;
+
+    let relay_kind: RelayKind = args
+        .relay_kind
+        .parse()
+        .map_err(|e| anyhow::anyhow!("Invalid relay kind: {}", e))?;
+
+    let cancel = CancellationToken::new();
+
+    // Read bootstrap peer if provided
+    let bootstrap_peers = if let Some(ref peer_file) = args.bootstrap_peer_file {
+        if peer_file.exists() {
+            info!("Reading bootstrap peer from {:?}", peer_file);
+            let content =
+                fs::read_to_string(peer_file).context("Failed to read bootstrap peer file")?;
+            let endpoint_addr: EndpointAddr = serde_json::from_str(&content)
+                .context("Failed to parse bootstrap peer endpoint")?;
+            info!("Bootstrap peer: {}", endpoint_addr.id.fmt_short());
+            vec![endpoint_addr]
+        } else {
+            info!("Bootstrap peer file not found yet, starting without peers");
+            vec![]
+        }
+    } else {
+        info!("No bootstrap peer file specified");
+        vec![]
+    };
+
+    info!("Initializing P2P network...");
+
+    let metrics = Arc::new(ClientMetrics::default());
+    let run_id = "inference";
+
+    type P2PNetwork = NetworkConnection<InferenceGossipMessage, ()>;
+
+    let mut network = P2PNetwork::init(
+        run_id,
+        None, // port (let OS choose)
+        None, // interface
+        discovery_mode,
+        relay_kind,
+        bootstrap_peers,
+        None, // secret key (generate new)
+        allowlist::AllowAll,
+        metrics.clone(),
+        Some(cancel.clone()),
+    )
+    .await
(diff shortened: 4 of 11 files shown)


psyche-book/src/explain/index.md
```diff
@@ -46,68 +48,71 @@ and model snapshots
...
-
-Once the _Witness_ phase concludes, the coordinator returns to the _Training_ phase. Clients are assigned new data, and the process repeats. After a predefined number of rounds, a _Cooldown_ round occurs, marking the end of an **epoch**.
+Once a witness quorum is reached, the coordinator advances to the _Training_ phase to allow all clients a brief window to download every training result of the previous round, clients are assigned new data, and the process repeats. After a predefined number of rounds, a _Cooldown_ round occurs, marking the end of an **epoch**.
 
 ## The witness/train loop visualized
 
 Here's a high-level overview of the process.
 
-Additional details exist, but this captures the overall flow of a single Round from an Epoch:
+There's additional implementation details, but this captures the overall flow of a single Round in an Epoch:
```
dsocolobsky:
```suggestion
There are additional implementation details, but this captures the overall flow of a single Round in an Epoch:
```

dsocolobsky:
I took a general look at the new docs, looks much better!

psyche-book/src/enduser/create-run.md
```diff
@@ -1,74 +1,94 @@
...
 
 #### Join Authority for Private Runs
 
-If we'll only allow some users to join the run we'll need to create one authorization per user (each user can then set multiple delegate keys later)
+If we'll only allow some users to join the run we'll need to create one authorization per user (each user can then set multiple delegate keys later) For example to use it locally we can do
 
 ```sh
-sh scripts/join-authorization-create.sh [RPC] join_authority.json [MY_USER_PUBKEY]
+just run_authorizer rpc=<RPC> grantee=<GRANTOR>
 ```
```
IAvecilla, line 50: @crypto-vincent I might need your help here. I'm trying to improve the docs for creating the runs and updating to the latest changes, the permissionless runs works correctly but I couldn't create a permissioned one.

Basically I'm doing:
1. Deploy of the coordinator and the authorizer
2. Running the authorizer script (now a just command) with: `sh scripts/join-authorization-create.sh localnet ~/.config/solana/id.json test.json` being the test.json the keypair of the user I want to allow to the run
3. Then following the create-run, update-config and set-paused commands explained in the docs (Not using a join authority since I think it will be used the payer by default.
4. Trying to join any client from there, even the main authority I'm getting an error saying that `Account is not created`.

I can write the docs not necessarily for you to take that but wanted to know if I'm overlooking something here. If you need the full error lmk, but maybe you already see something's wrong

architectures/decentralized/solana-authorizer/README.md
```diff
@@ -49,7 +49,7 @@ impl Authorization {
 
 The smart contract then exposes a set of instruction to manipulate those `Authorization` PDAs:
 
-- `authoziation_create`, create a new PDA
+- `authorization_create`, create a new PDA
```
crypto-vincent, line 52: nice catch!

psyche-book/src/enduser/create-run.md
crypto-vincent: Got ya, it sounds like it's probably a small missmatch between the authorization being created and the authorization passed as parameter during joining. Hard to tell without understanding the whole flow of what's going on here but:

I see that by default the "grantee" is the "1111...1111" address, so that authorization's PDA should also be passed as the authorization for each joining client during the join.

Im going to try to read the chain of events from the scripts, but im not yet familiar with that part of the code

psyche-book/src/enduser/create-run.md
```diff
@@ -1,76 +1,96 @@
...
+We'll need a key-pair file that manages join permissions, it can be the default created by Solana when you do `solana-keygen new` located in `~/.config/solana/id.json`
 
 #### Join Authority for Public Runs
 
 If we're looking to make a permissionless run (anyone can join), we'll need to create an authorization that's valid for everyone.
 
+Running:
+
+```sh
+just run_authorizer
```
crypto-vincent: nit: might want to add the rpc=<value> here since most of the time it's probably not going to be run on localnet by users

psyche-book/src/enduser/create-run.md
```diff
@@ -1,76 +1,96 @@
...
 ```
 
 #### Join Authority for Private Runs
 
-If we'll only allow some users to join the run we'll need to create one authorization per user (each user can then set multiple delegate keys later)
+If we'll only allow some users to join the run we'll need to create one authorization per user (each user can then set multiple delegate keys later) For example to use it locally we can do
 
 ```sh
-sh scripts/join-authorization-create.sh [RPC] join_authority.json [MY_USER_PUBKEY]
+just run_authorizer rpc=<RPC> grantee=<GRANTOR>
```
crypto-vincent: there seem to be a typo here: `grantee=<GRANTOR>` most likely both of those values could be passed as parameter, there's no `join_authority.json` involved in here ? im a bit confused

psyche-book/src/enduser/create-run.md
```diff
@@ -1,76 +1,96 @@
...
+
+- `rpc`: The RPC URL to use for the Solana network. By default: `http://127.0.0.1:8899`
+- `grantor`: The path to the file with a Solana Keypair, will be used to create the authorization and grant access to the run. By default: `~/.config/solana/id.json`
+- `grantee`: The public key of the user that will be granted access to the run. By default: `11111111111111111111111111111111` that means is permissionless.
+
+You can override any of these values like this:
+
 ```sh
-sh scripts/join-authorization-create.sh [RPC] join_authority.json 11111111111111111111111111111111
+just rpc=<value> grantor=<value> grantee=<value> run_authorizer
```
crypto-vincent: im not too familiar with `just` but can the parameter be place both before and after the command ?

psyche-book/src/enduser/create-run.md
crypto-vincent: For sanity purposes: in order for the system to work during join:
- the `GRANTOR` needs to be the `join_authority`
- the `GRANTEE` needs to be the joining client (or 1111...1111)

In order to diagnose this we'd need:
 - 1) the command run during creation of the authorization
 - 2) the command run for the run-creation
 - 3) the commadn run for the client joining
 
 With those 3 we should be able to figure it out

IAvecilla: Yeah, I think I'm doing all the steps correctly but is not working for me, this is the exact commands I'm running in case you want to reproduce it or to follow the exact same steps.

1. I run a validator with `solana-test-validator -r`
2. I run `./scripts/deploy-solana-test.sh` that basically deploys the coordinator and the authorizer in the local validator.
3. I create a new keypair with `solana-keygen new -o test.json` that will be my grantee, is the one that I want to authorize into the run.
4. Run `just grantee=$(solana-keygen pubkey test.json) run_authorizer`. This will run the `join-authorization-create.sh` script with the grantee being the pubkey of the new `test.json` keypair created and the grantor being the wallet on `~/.config/solana/id.json` that is the one used by default in all the scripts.
5. Then create the run with: 
```
cargo run --bin psyche-solana-client -- create-run \
    --rpc http://127.0.0.1:8899
    --run-id test \
    --join-authority $(solana-keygen pubkey) \ 
    --wallet-private-key-path ~/.config/solana/test.json \
    --client-version "latest"
```
This should use the join authority `~/.config/solana/id.json` if I understand correctly
6. Update config and set-paused:
```
cargo run --bin psyche-solana-client -- update-config \
    --rpc http://127.0.0.1:8899 \
    --run-id test \
    --config-path ./config/solana/light-config.toml \
    --wallet-private-key-path ~/.config/solana/id.json

cargo run --bin psyche-solana-client -- set-paused \
    --rpc http://127.0.0.1:8899 \
    --run-id test \
    --resume \
    --wallet-private-key-path ~/.config/solana/id.json
```
7. Try to join the grantee client with:
```
WALLET_FILE=test.json just dev start-training-localnet-light-client
```
That it just calls:
```
    cargo run --release --bin psyche-solana-client -- \
        train \
        --wallet-private-key-path ./test.json \
        --rpc  http://127.0.0.1:8899 \
        --ws-rpc ws://127.0.0.1:8900 \
        --run-id test \
        --data-parallelism 1 \
        --tensor-parallelism 1 \
        --micro-batch-size 8 \
        --logs "console" \
        "$@"
```

And with that I get the `AccountNotInitialized` errors. Also if I use `WALLET_FILE=~/.config/solana/id.json` that is the run creator it also fails with the same error

psyche-book/src/enduser/create-run.md
IAvecilla: Good idea, changed it!

psyche-book/src/enduser/create-run.md
IAvecilla: Yeah, I was waiting to get the permissioned run creation to work to think on how to write that section so it was wrong and outdated. I changed it to have at least a correct placeholder but I'll change it in the near future explaining how all that works

IAvecilla: Yes, the just command basically executes that script but it is more user friendly to give a just command in my opinion instead of running the script directly 
```
run_authorizer:
    sh scripts/join-authorization-create.sh {{ rpc }} {{ grantor }} {{ grantee }}
```

psyche-book/src/enduser/create-run.md
IAvecilla: Yes you can but if you use it that way you can't use default values, you always have to execute the just command with all the parameters in the correct position. This way you can override the parameter that you need and the others will take the default values.

psyche-book/src/enduser/create-run.md
crypto-vincent: On a high level, this sounds like it is indeed the right chain of events. Im a bit confused with each exact wallet being used in all context, for example im assuming that:
- `test.json` is the same wallet as `~/.config/solana/test.json`
- `$(solana-keygen pubkey)` is the same wallet as `~/.config/solana/id.json`

I think it's just easier in this case to look at the logs. 

The script `join-authorization-create.sh` will log a few useful variables:

```sh
echo "GRANTOR_PUBKEY: $GRANTOR_PUBKEY"
echo "GRANTEE_PUBKEY: $GRANTEE_PUBKEY"
echo "AUTHORIZATION_PUBKEY: $AUTHORIZATION_PUBKEY"
```

Finding those logs and printing the exact authorization PDA being passed as parameter to the "join" coordinator instruction would probably give us the answer.

crypto-vincent: Re-reading this, I think this is the issue:

```sh
just grantee=$(solana-keygen pubkey test.json) run_authorizer
```

This means that you're creating an authorization that is ONLY valid for the grantee `test.json`. (I think before it was creating an authorization valid for everyone using `1111...1111` as grantee)

If doing so, you'll need to pass the `--authorizer $(solana-keygen pubkey test.json)` flag to the train/join coordinator CLI instructions.

This is because the `--authorizer` flag defaults to `1111...1111` for historical simplicity.

psyche-book/src/enduser/create-run.md
crypto-vincent: Got it, but in this case, shouldnt the "join_authority" be passed as the "grantor" parameter ?

psyche-book/src/enduser/create-run.md
crypto-vincent: Oh ok, interesting, so putting the parameters before the command name is "more explicit" ? if so, we should give all our examples in that way maybe ? Up to you

psyche-book/src/enduser/create-run.md
IAvecilla: Okay, that worked! I didn’t know that parameter existed, my bad. Also, sorry for the confusion: there was an error in the commands, and instead of `~/.config/solana/test.json` it should have been `~/.config/solana/id.json`.

That said, this seems to solve the issue. I’ll try to adapt the script so it can easily use a non-default authorizer as well. Thank you!

psyche-book/src/enduser/create-run.md
IAvecilla: With the latest changes on the authorization I just ended removing that command and just use the new cli I think it's cleaner

PR #423, opened 2025-12-09
Adding vLLM Inference with UpdatableLLMEngine

samherring99:
This PR encapsulates the following changes to get vLLM working in the Psyche environment:

1. `engine.py`

  - UpdatableLLMEngine class wrapping vLLM's LLMEngine
  - Configurable parameters: model name, tensor parallelism, dtype, max sequence length, GPU memory utilization
  - Core methods:
    - `add_request()` - Queue inference request with sampling parameters
    - `step()` - Process one inference step (continuous batching)
    - `has_unfinished_requests()` - Check for pending requests
    - `get_tokenizer()` - Access tokenizer for the loaded model
    - `shutdown()` - Clean shutdown with Ray cleanup

The `UpdatableLLMEngine` class is designed to account for dynamic / in-place weight updates in the future, as right now it operates by creating a new instance of the Engine in a separate process to reload weights.

Setup:
```
nix develop .#dev-python
python -m venv .venv
source .venv/bin/activate

uv pip install vllm
uv pip install -e python
```

Local tests passing:

Screenshot 2025-12-09 at 11 25 46 AM

### DIFF
--- a/python/python/psyche/vllm/engine.py
+++ b/python/python/psyche/vllm/engine.py
@@ -1,0 +1,85 @@
+import logging
+from itertools import count
+from typing import List, Optional, Dict, Any
+
+logger = logging.getLogger(__name__)
+
+try:
+    from vllm import LLMEngine, EngineArgs, SamplingParams, RequestOutput
+
+    VLLM_AVAILABLE = True
+except ImportError:
+    VLLM_AVAILABLE = False
+    logger.warning("vLLM is not installed. UpdatableLLMEngine will not be available.")
+    LLMEngine = Any
+    EngineArgs = Any
+    SamplingParams = Any
+    RequestOutput = Any
+
+
+# A wrapper around vLLM's LLMEngine that supports dynamic weight updates.
+class UpdatableLLMEngine:
+    def __init__(
+        self,
+        model_name: str,
+        tensor_parallel_size: int = 1,
+        dtype: str = "auto",
+        max_model_len: Optional[int] = None,
+        gpu_memory_utilization: float = 0.90,
+    ):
+        if not VLLM_AVAILABLE:
+            raise ImportError(
+                "vLLM is not installed. Please install it with: pip install vllm"
+            )
+
+        self.model_name = model_name
+        self.tensor_parallel_size = tensor_parallel_size
+        self.dtype = dtype
+        self.max_model_len = max_model_len
+        self.gpu_memory_utilization = gpu_memory_utilization
+
+        logger.info(f"Initializing vLLM engine with model: {model_name}")
+
+        engine_args = EngineArgs(
+            model=model_name,
+            tensor_parallel_size=tensor_parallel_size,
+            dtype=dtype,
+            max_model_len=max_model_len,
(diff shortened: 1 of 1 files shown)


python/python/psyche/vllm/engine.py
```diff
@@ -0,0 +1,86 @@
...
+    def abort_request(self, request_id: str):
+        self.engine.abort_request(request_id)
+
+    def get_tokenizer(self):
+        return self.engine.tokenizer
+
+    def shutdown(self):
+        logger.info("Shutting down vLLM engine")
+        # vLLM handles cleanup automatically when the engine is destroyed
+        del self.engine
```
ethernet8023, line 84: this line and `     self.engine = None` are functionally identical - both remove the reference, though `del self.engine` deletes the property entirely and `= None` just sets it to `None`. I'd remove one of these two.

nix/nixpkgs.nix
```diff
@@ -17,7 +17,25 @@ in
       (final: prev: {
         python312Packages = prev.python312Packages.override {
           overrides = pyfinal: pyprev: {
-            torch = pyfinal.torch-bin;
+            torch =
+              (if pyprev ? torch-bin_2_9 then pyprev.torch-bin_2_9 else pyprev.torch-bin).overrideAttrs
```
ethernet8023: Why torch-bin_2_9? the torch-bin in nixpkgs we already use is 2.9.0 to my knowledge.

nix/nixpkgs.nix
```diff
@@ -17,7 +17,25 @@ in
...
+                (old: {
+                  passthru = (old.passthru or { }) // {
+                    cudaSupport = cudaSupported;
+                    rocmSupport = false;
+                    cudaPackages = final.cudaPackages;
+                    cudaCapabilities =
```
ethernet8023: The CUDA compute specifiers also mean that we won't be able to use *any* CUDA features on GB200,B200, and higher, right? and also no 8.6 so no RTX 30xx, etc. My understanding is that the default compute levels are *all of them*, and we shouldn't need to decrease that list.

arilotter (changes requested):

nix/nixpkgs.nix
samherring99: 8.6 is a good call, but I have been developing this on B200s and it's working, I'll pull it all out and see if this might be overkill

nix/nixpkgs.nix
samherring99: I'd left these in from getting vLLM to work previously due to some weird import errors with uv. I'll also try reverting this and see what happens, will update here.

nix/nixpkgs.nix
samherring99: Ok yeah this comment, and the one above, this was leftover from trying to install vllm through nix store, it works fine when installed with uv :-) will revert.

nix/nixpkgs.nix
samherring99: See below comment, reverting this :-)

arilotter (approved):
actually getting vllm into the repo is left as an exercise for me at a later date

PR #431, opened 2025-12-12
Rust Bridge For Psyche Inference Node

samherring99:
## Description
This PR adds the following files to continue the Psyche Inference Node implementation:

- `python/python/psyche/rust_bridge.py` - a Python bridge wrapper to allow PyO3 to call the UpdatableLLMEngine methods from Rust calls. This file exposes the following methods:
  - `create_engine` - creates an UpdatableLLMEngine
  - `run_inference` - inferences with an existing UpdatableLLMEngine
  - `shutdown_engine` - shuts down the specified engine
  - `get_engine_stats` - returns engine statistics like ("status", "engine_id", "model_name", "tensor_parallel_size","has_unfinished_requests")
  - `list_engines` - lists available UpdatableLLMEngine instances

- `python/python/psyche/run_inference_subprocess.py` - a helper file to allow inference through the UpdatableLLMEngine to be run as a subprocess (for potential reloading in the future)

- `python/python/extension-impl/src/vllm.rs` - Rust implementation using PyO3 to access methods exposed in `rust_bridge.py`. This file matches all the method exposed in the Python<->Rust bridge and provides their Rust counterparts.

- `python/python/extension-impl/tests/test_vllm_integration.rs` - Rust test for vLLM integration, tests creation and shutdown of engine, plus `list_engines` call from Rust.

- `python/python/extension-impl/tests/test_subprocess_reload.rs` - Rust test for spawning subprocesses from Rust which allows for checkpoint reloading. 

## NOTE
This code is set up to allow for easy reloading with future RL integrations, which is why I'm setting it up using this reloadable subprocess implementation. For the standard inference-only use case this shouldn't have any real impact, so I'm open to discussion on if we want to remove this and add back in the future when we integrate RL into Psyche.

## Tests

To setup the environment and run the tests:
```
nix develop .#dev-python
uv venv

source .venv/bin/activate
uv pip install vllm

cargo build -p psyche-python-extension --release
cp target/release/libpsyche_python_extension.so python/python/psyche/_psyche_ext.so

uv pip install -e python

cargo test -p psyche-python-extension-impl
```

### DIFF
--- a/python/extension-impl/src/vllm.rs
+++ b/python/extension-impl/src/vllm.rs
@@ -22,6 +22,77 @@ use pyo3::prelude::*;
 use pyo3::types::PyDict;
 use std::collections::HashMap;
 
+/// Response from engine creation
+#[derive(Debug, Clone)]
+pub struct EngineCreationResult {
+    pub success: bool,
+    pub engine_id: Option<String>,
+    pub error: Option<String>,
+}
+
+/// Response from inference request
+#[derive(Debug, Clone)]
+pub struct InferenceResult {
+    pub success: bool,
+    pub request_id: Option<String>,
+    pub generated_text: Option<String>,
+    pub full_text: Option<String>,
+    pub prompt: Option<String>,
+    pub error: Option<String>,
+}
+
+/// Response from engine shutdown
+#[derive(Debug, Clone)]
+pub struct ShutdownResult {
+    pub success: bool,
+    pub engine_id: Option<String>,
+    pub error: Option<String>,
+}
+
+/// Response from engine stats request
+#[derive(Debug, Clone)]
+pub struct EngineStats {
+    pub success: bool,
+    pub engine_id: Option<String>,
+    pub model_name: Option<String>,
+    pub tensor_parallel_size: Option<i64>,
+    pub has_unfinished_requests: Option<bool>,
+    pub error: Option<String>,
+}
+
+/// Response from list engines request
+#[derive(Debug, Clone)]
+pub struct EngineList {
+    pub success: bool,
+    pub engine_ids: Vec<String>,
+    pub error: Option<String>,
+}
+
+/// Helper function to convert Python dict to Rust HashMap
+fn py_dict_to_hashmap(dict: &Bound<'_, PyDict>) -> PyResult<HashMap<String, PyObject>> {
+    let mut map = HashMap::new();
+    for (key, value) in dict.iter() {
+        let key_str: String = key.extract()?;
+        map.insert(key_str, value.unbind());
+    }
+    Ok(map)
+}
+
+/// Helper to extract optional string from HashMap
+fn get_optional_string(map: &HashMap<String, PyObject>, key: &str, py: Python) -> Option<String> {
+    map.get(key).and_then(|v| v.extract(py).ok())
+}
+
+/// Helper to extract optional i64 from HashMap
+fn get_optional_i64(map: &HashMap<String, PyObject>, key: &str, py: Python) -> Option<i64> {
+    map.get(key).and_then(|v| v.extract(py).ok())
+}
+
+/// Helper to extract optional bool from HashMap
+fn get_optional_bool(map: &HashMap<String, PyObject>, key: &str, py: Python) -> Option<bool> {
+    map.get(key).and_then(|v| v.extract(py).ok())
+}
+
 /// Create a new vLLM engine
 pub fn create_engine(
     py: Python,
@@ -31,7 +102,7 @@ pub fn create_engine(
     dtype: Option<&str>,
     max_model_len: Option<i32>,
     gpu_memory_utilization: Option<f64>,
-) -> PyResult<HashMap<String, PyObject>> {
+) -> PyResult<EngineCreationResult> {
     let rust_bridge = py.import("psyche.vllm.rust_bridge")?;
 
     let kwargs = PyDict::new(py);
@@ -55,17 +126,17 @@ pub fn create_engine(
     }
 
     let result = rust_bridge.call_method("create_engine", (), Some(&kwargs))?;
-
-    // Convert Python dict to Rust HashMap
     let dict = result.downcast::<PyDict>()?;
-    let mut map = HashMap::new();
+    let map = py_dict_to_hashmap(dict)?;
 
-    for (key, value) in dict.iter() {
-        let key_str: String = key.extract()?;
-        map.insert(key_str, value.unbind());
-    }
+    let status = get_optional_string(&map, "status", py).unwrap_or_default();
+    let success = status == "success";
 
-    Ok(map)
+    Ok(EngineCreationResult {
+        success,
+        engine_id: get_optional_string(&map, "engine_id", py),
+        error: get_optional_string(&map, "error", py),
+    })
 }
 
 /// Run inference on an engine
@@ -76,7 +147,7 @@ pub fn run_inference(
     temperature: Option<f64>,
     top_p: Option<f64>,
     max_tokens: Option<i32>,
-) -> PyResult<HashMap<String, PyObject>> {
+) -> PyResult<InferenceResult> {
     let rust_bridge = py.import("psyche.vllm.rust_bridge")?;
 
     let kwargs = PyDict::new(py);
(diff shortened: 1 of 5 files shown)


python/extension-impl/src/vllm.rs
```diff
@@ -0,0 +1,178 @@
...
+    let result = rust_bridge.call_method("create_engine", (), Some(&kwargs))?;
+
+    // Convert Python dict to Rust HashMap
+    let dict = result.downcast::<PyDict>()?;
+    let mut map = HashMap::new();
+
+    for (key, value) in dict.iter() {
+        let key_str: String = key.extract()?;
+        map.insert(key_str, value.unbind());
+    }
```
IAvecilla: I think this pattern repeats a lot, maybe we can extract it to an aux function that `py_dict_to_hashmap` or something similar

python/extension-impl/src/vllm.rs
```diff
@@ -0,0 +1,178 @@
...
+/// Create a new vLLM engine
+pub fn create_engine(
+    py: Python,
+    engine_id: &str,
+    model_name: &str,
+    tensor_parallel_size: Option<i32>,
+    dtype: Option<&str>,
+    max_model_len: Option<i32>,
+    gpu_memory_utilization: Option<f64>,
+) -> PyResult<HashMap<String, PyObject>> {
```
IAvecilla: Maybe it would be good to introduce a new struct, `VLLMEngine` (or something similar), and return that instead. You could implement `From<PyDict>` or `From<HashMap>` and construct it directly from the map. That way, it would be self-evident which fields the result contains, instead of having to print the `HashMap` to see which attributes are available.

python/extension-impl/src/vllm.rs
```diff
@@ -0,0 +1,178 @@
...
+
+/// Run inference on an engine
+pub fn run_inference(
+    py: Python,
+    engine_id: &str,
+    prompt: &str,
+    temperature: Option<f64>,
+    top_p: Option<f64>,
+    max_tokens: Option<i32>,
+) -> PyResult<HashMap<String, PyObject>> {
```
IAvecilla: Same as `create_engine` suggestion

python/extension-impl/src/vllm.rs
IAvecilla: Maybe we could even use an `EngineBuilder` or something similar and avoid calling `create_engine` with all the optional fields set to `None`. Instead, we could do something like `EngineBuilder::create_engine(py, engine_id, model_name).with_tensor_parallel_size(5).build()`. But that’s just a preference, there’s nothing wrong with the current approach.

python/python/psyche/vllm/engine.py
```diff
@@ -81,5 +81,22 @@ def get_tokenizer(self):
...
+
+            if ray.is_initialized():
+                logger.info("Shutting down Ray...")
+                ray.shutdown()
+                logger.info("Ray shutdown complete")
+        except Exception as e:
+            logger.warning(f"Failed to shutdown Ray: {e}")
+
+        # Give time to clean up
+        import time
```
IAvecilla: We can add this import to the top with all the other imports

python/python/psyche/vllm/run_inference_subprocess.py
```diff
@@ -0,0 +1,75 @@
...
+
+os.environ["VLLM_LOGGING_LEVEL"] = "ERROR"
+
+multiprocessing.set_start_method("spawn", force=True)
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent))
+
+
+def run_inference(model_name: str, prompt: str) -> str:
+    import contextlib
```
IAvecilla: I think this lib is not being used

python/python/psyche/vllm/run_inference_subprocess.py
```diff
@@ -0,0 +1,75 @@
...
+                max_model_len=512,
+                gpu_memory_utilization=0.3,
+            )
+
+            sampling_params = {
+                "temperature": 0.0,  # Deterministic
+                "max_tokens": 20,
+            }
+
+            request_id = engine.add_request(prompt, sampling_params)
```
IAvecilla, line 39: This variable is not being used

python/python/psyche/vllm/run_inference_subprocess.py
```diff
@@ -0,0 +1,75 @@
...
+            outputs = []
+            while engine.has_unfinished_requests():
+                outputs.extend(engine.step())
+
+            generated_text = outputs[0].outputs[0].text
```
IAvecilla: Should we check that the output is not empty? I think this will crash if that's the case

python/python/psyche/vllm/run_inference_subprocess.py
```diff
@@ -0,0 +1,75 @@
...
+
+            engine.shutdown()
```
IAvecilla: Maybe we should add this to the `finally` clause? I don't exactly know if it's possible but if something goes wrong but we created the engine we're not shutting it down

python/python/psyche/vllm/run_inference_subprocess.py
```diff
@@ -0,0 +1,75 @@
...
+
+    try:
+        generated = run_inference(model_name, prompt)
+        import json
+
+        result = {"generated_text": generated}
+        print(json.dumps(result))
+    except Exception as e:
+        print(f"Error: {e}", file=sys.stderr)
+        import traceback
```
IAvecilla: I would suggest adding this imports at the top, but it's personal preference

python/python/psyche/vllm/rust_bridge.py
```diff
@@ -0,0 +1,137 @@
+import logging
+from typing import Dict, Any, Optional
+from pathlib import Path
```
IAvecilla: I think this is not being used

python/python/psyche/vllm/rust_bridge.py
```diff
@@ -0,0 +1,137 @@
...
+
+        engine = _engines[engine_id]
+
+        sampling_params = {
+            "temperature": temperature,
+            "top_p": top_p,
+            "max_tokens": max_tokens,
+        }
+
+        request_id = engine.add_request(prompt, sampling_params)
```
IAvecilla, line 64: Also not being used

python/python/psyche/vllm/rust_bridge.py
```diff
@@ -0,0 +1,137 @@
...
+    top_p: float = 1.0,
+    max_tokens: int = 100,
+) -> Dict[str, Any]:
+    try:
+        if engine_id not in _engines:
+            error_msg = f"Engine '{engine_id}' not found"
+            logger.error(error_msg)
+            return {"status": "error", "error": error_msg}
+
+        engine = _engines[engine_id]
```
IAvecilla, line 56: We're not checking if the `engine_id` is already present in the map so we could be overriding it here, maybe we should check before adding it.

python/python/psyche/vllm/rust_bridge.py
```diff
@@ -0,0 +1,137 @@
+import logging
+from typing import Dict, Any, Optional
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+# Global registry of engines by ID
+_engines: Dict[str, Any] = {}
```
IAvecilla, line 7: I have a question here, is this expected to run concurrently, like many threads could call `create_engine` and `run_inference` if that's the case we should add some sort of lock or concurrency control here because we could get some race conditions and unexpected behavior using this global Dictionary. If we are sure that we will only call it once it should be okay

python/python/psyche/vllm/run_inference_subprocess.py
samherring99: good catch, forgot to remove :-)

python/python/psyche/vllm/run_inference_subprocess.py
samherring99: will do for this and all the above imports

python/python/psyche/vllm/rust_bridge.py
samherring99: also forgot to remove, ty :-)

python/python/psyche/vllm/rust_bridge.py
samherring99: This idea was more along the lines of one instance existing at a time, but keeping a way to store and track an engine that is shutting down vs. an engine that is starting up with a new checkpoint ( for RL later on ) - I'm open to refactoring this if we feel like it's not necessary, I don't love global variables 🙃

python/python/psyche/vllm/rust_bridge.py
samherring99: I meant to include in return body, I think that would make sense, yeah?

IAvecilla: Yeah! that sounds reasonable

IAvecilla:
I was able to run the python env and it seems that the testing commands in the PR description work as expected. Running `cargo test -p psyche-python-extension-impl --features vllm-tests` to run the tests you added in `test_vllm_integration.rs` and `test_subprocess_reload.rs` I get a few errors. Is there a specific setup that I need to run them?

python/python/psyche/vllm/run_inference_subprocess.py
samherring99: This one is okay to me, it's a helper function for tests and isn't planned to be used in production, which is why it's set up this way

jquesnelle:
> we resolved this to being a dependency issue that will be fixed in ([main...vlllm](https://github.com/PsycheFoundation/psyche/compare/main...vlllm)) - so this is non-blocking

do we need the `vlllm` branch to test this?

python/python/psyche/vllm/rust_bridge.py
IAvecilla: Okay yeah, if that's the case maybe we can keep it as it is

PR #435, opened 2025-12-12
torchtitan

jquesnelle:


python/python/psyche/models/hf_transformers.py
```diff
@@ -78,6 +78,29 @@ def build_mesh(device_type, pp=1, dp_replicate=1, dp_shard=1, cp=1, tp=1) -> Dev
...
+
+    config = config_class.from_dict(config)
+
+    # If n_routed_experts is None disable all MoE related features
+    if hasattr(config, "n_routed_experts") and config.n_routed_experts is None:
+        config.n_shared_experts = None
+        config.moe_layer_freq = None
+        config.num_experts_per_tok = None
+        config.moe_intermediate_size = None
+        # Set first_k_dense_replace to a very large number to force all layers to be dense
```
ethernet8023: silly nit - i doubt `num_hidden_layers` will ever be > one million... 
but rather than an arbitrary large number, do you think setting this to `num_hidden_layers` (or +1) would signify the intention more clearly?

feel free to ignore

python/python/psyche/models/hf_transformers.py
```diff
@@ -146,8 +149,7 @@ def from_pretrained(
                 config,
                 attn_implementation=attn_implementation,
             )
-        if device.type == "cuda":
-            torch.cuda.set_device(device)
+        torch.cuda.set_device(device)
```
ethernet8023, line 141: does this mean that the torchtitan code can't run on CPU, e.g. for tests?

python/python/psyche/models/ttitan.py
```diff
@@ -0,0 +1,472 @@
...
+        self.loss_fn = loss_fn
+        self.config = config
+        self.config_tt = config_tt
+        self.job_config = job_config
+        self.device = device
+        self.amp = amp
+        self.parallel_dims = parallel_dims
+
+    @staticmethod
+    def convert_config(config, override_max_position_embeddings: Optional[int] = None):
```
ethernet8023, line 66: i wonder if we can build some registry of model types to pull this from, just so they don't all have to sit in a big if/else, and we could instead throw them in some `configs` folder as separate modules. not a blocker by any means though

python/python/psyche/models/ttitan.py
```diff
@@ -0,0 +1,472 @@
...
+                head_dim=config.head_dim,
+                hidden_dim=config.intermediate_size,
+                norm_eps=config.rms_norm_eps,
+                rope_theta=config.rope_theta,
+                max_seq_len=seq_len,
+                moe_enabled=True,
+                moe_inter_dim=config.moe_intermediate_size,
+                moe_args=MoEArgs(
+                    num_experts=config.num_experts,
+                    num_shared_experts=0,  # qwen3 has no shared experts
```
ethernet8023, line 124: should we throw an error if the config has shared experts set?

python/python/psyche/models/ttitan.py
```diff
@@ -0,0 +1,472 @@
...
+                    top_k=config.num_experts_per_tok,
+                    score_func="softmax",
```
ethernet8023, line 126: should we throw an error if the config has `config.scoring_func` set?

python/python/psyche/models/ttitan.py
```diff
@@ -0,0 +1,472 @@
...
+                    "original_max_position_embeddings"
+                ],
+                rope_theta=config.rope_theta,
+                rope_factor=config.rope_scaling["factor"],
+                beta_fast=config.rope_scaling["beta_fast"],
+                beta_slow=config.rope_scaling["beta_slow"],
+                moe_args=MoEArgs(
+                    num_experts=config.num_local_experts,
+                    num_shared_experts=0,
+                    score_func="softmax",
```
ethernet8023, line 189: same thoughts as above - should we error if these are set in the config?

python/python/psyche/models/ttitan.py
```diff
@@ -0,0 +1,472 @@
...
+                decoder_sparse_step=config.decoder_sparse_step,
+                full_attention_interval=config.full_attention_interval,
+                linear_num_key_heads=config.linear_num_key_heads,
+                linear_num_value_heads=config.linear_num_value_heads,
+                linear_key_head_dim=config.linear_key_head_dim,
+                linear_value_head_dim=config.linear_value_head_dim,
+                linear_conv_kernel_dim=config.linear_conv_kernel_dim,
+                moe_args=MoEArgs(
+                    num_experts=config.num_experts,
+                    num_shared_experts=1,  # constant?
```
ethernet8023, line 220: same as above, should we error if set in config?

python/python/psyche/models/ttitan.py
```diff
@@ -0,0 +1,472 @@
...
+                    top_k=config.num_experts_per_tok,
+                    score_func="softmax",
```
ethernet8023, line 222: same as above, should we error if set in config?

python/python/psyche/models/hf_transformers.py
jquesnelle: the device could be "cpu", unsure why we had this conditional ever

python/python/psyche/models/ttitan.py
jquesnelle: it's that `config` (the HF one) has no variable for `shared_experts` for Qwen3 but TT uses `MoEArgs` for _all_ MoE implementations (DSV3, Q3, L4, etc.) so it's like the superset of all configurations. For the missing ones we need to set it to whatever the HF one implicitly has

jquesnelle:
upgraded by https://explorer.solana.com/tx/Zo78tt5r9bZsfYf1oVk5fUq5bepzHZsQrgGhLrcQoVznpvGogUC9dyXCTr5RbMR3iK52bKzpCLDeMfkeZ77ziMw?cluster=devnet

PR #454, opened 2025-12-23
evals: arc_easy improvements

### LINKED ISSUE #446: Evals: Fix Fewshots
We need to do a double check in our evalutions with fewshots.

1. We are tokenizing the fewshots and the question separatly and the extendinf the String
https://github.com/PsycheFoundation/psyche/blob/main/shared/eval/src/harness.rs#L135
https://github.com/PsycheFoundation/psyche/blob/main/shared/eval/src/harness.rs#L148

That can lead to some mismatch in the tokens generations
For example in our implementation of arc easy and arc challenge, if the eval has fewshots, if we decode the eval request we have and additional space
' Answer: ....'

You can see the Hugging Face request to compare here `batched_inps`:
https://github.com/EleutherAI/lm-evaluation-harness/blob/cd9bac7c27f3c876bb8e60dca8ee3b6de6b33c35/lm_eval/models/huggingface.py#L1296-L1297

2. I think here we need an extra space 
https://github.com/PsycheFoundation/psyche/blob/abb433574089f6942b5e4141935c65d2a3e03c99/shared/eval/src/harness.rs#L243-L244

We should check all the evals and see that our implementation has the same format as `lm_eval`

dsocolobsky:
wip

we detected some differences with how we tokenize stuff in the evals code specially regarding arc_easy task, this attempts to improve it

1. Tokenize fewshot prefixes + question text together
2. Small hack to avoid an extra space appearing when we tokenized
3. Do not shuffle fewshot samples as lm_evals seems not to do that
4. Use training split for fewshot samples instead of test, as lm_eval seems to do that.

In general it seems to match more closely lm_evals' results

```
lm_evals
| Tasks  |Version|Filter|n-shot| Metric |   |Value |   |Stderr|
|--------|------:|------|-----:|--------|---|-----:|---|-----:|
|arc_easy|      1|none  |     3|acc     |↑  |0.6747|±  |0.0096|
|        |       |none  |     3|acc_norm|↑  |0.6578|±  |0.0097|


main
ARC-Easy: {"acc_norm": 0.6203703703703703, "acc": 0.6611952861952862}

branch
ARC-Easy: {"acc": 0.6717171717171717, "acc_norm": 0.6582491582491582}
```

closes #446


shared/eval/src/harness.rs
```diff
@@ -121,63 +121,70 @@ pub struct TokenizedGenerateUntilDocument {
...
-                    .collect::<Vec<_>>();
-                let choice_str = tokenizer.decode(choice_tokens, false).unwrap();
-                if choice_str.contains(choice) {
-                    let choice_tokens = choice_tokens.iter().map(|x| *x as i64).collect::<Vec<_>>();
-                    choices.push(choice_tokens.clone());
-                    choices_token_len.push(choice_tokens.len());
-                    break;
-                }
+            // Debug code to write tokens to file so we can compare
+            if let Ok(mut file) = std::fs::File::create("tokens.my_impl.out") {
```
pefontana: remove

shared/eval/src/harness.rs
```diff
@@ -259,7 +272,8 @@ impl Task {
             }
             TaskType::GenerateUntil(gu_docs) => {
                 let mut docs = gu_docs.get_documents();
-                docs.shuffle(&mut self.rand);
+                // I believe to match lm_eval we should not shuffle. But leaving this here just in case we want to try again shuffling
```
pefontana: Is necessary to keep this comment?

dsocolobsky: I think I would rather leave the comment to document we were shuffling before and not anymore, just in case I made a mistake.

shared/eval/src/harness.rs
```diff
@@ -236,11 +237,16 @@ impl Task {
...
                                 });
+                            // Build fewshots to match how test question is tokenized:
+                            // text (ends with "Answer:") + " " + choice
                             fewshot_examples.shuffle(&mut self.rand);
                             fewshot_examples
                                 .into_iter()
                                 .take(self.num_fewshot)
-                                .map(|x| format!("{}{}", x.text, x.choices[x.answer]))
+                                .map(|x| {
+                                    // Use same format as test question to ensure consistent tokenization
```
pefontana: I think this comment is really unnecessary

shared/eval/src/harness.rs
pefontana: I think we can create an issue for this.
This comments dont help in the codebase

PR #455, opened 2025-12-23
Cooldown state rework

IAvecilla:
Depends on #463

This PR contains several changes to client arguments, dependencies, and the general cooldown state logic. Here’s a summary of the changes made in this PR:

* All clients are now potential checkpointers and must be prepared to perform that task. At the end of an epoch, during the Cooldown state, a subset of clients (1/3 of the total clients, capped at `SOLANA_MAX_NUM_CHECKPOINTERS`, which is currently 16) is deterministically selected as checkpointers using a seeded shuffle algorithm based on the round’s random seed.
* Cooldown is now an opportunistic state, just like the others. Once a client uploads the entire model to external storage, it sends a transaction that moves the run forward to the next epoch. If no transaction is received, the coordinator waits for the full `cooldown_time` before moving on.
* Added cancellation support so that once one client successfully checkpoints, the others can cancel their upload tasks and stop uploading in order to move to the next epoch.
* The `--hub-repo` and `--gcs-bucket` flags were removed from the training arguments. Checkpoint destinations are now derived from the coordinator configuration, and there is a single repo or bucket that contains all the checkpoints for the run.
* Added a new client flag, `--skip-checkpoint-upload`, for testing purposes. This avoids uploading and saving checkpoints locally and should not be used outside testing environments.
* Updated the `google-cloud-storage` crate to version 1.6.
* Added bucket and hub repo permission verification before joining a run to ensure that the client can eventually checkpoint.
* The run manager now expects the GCP bucket credentials to be present inside the `SCRATCH_DIR` in order to work correctly when using GCP checkpoints.
* Updated the psyche book with an explanation of the new cooldown behavior.

### DIFF
--- a/architectures/decentralized/solana-client/src/app.rs
+++ b/architectures/decentralized/solana-client/src/app.rs
@@ -15,7 +15,7 @@ use psyche_client::{
 };
 use psyche_coordinator::{
     ClientState, Coordinator, CoordinatorError, RunState,
-    model::{self, CheckpointSource, Model},
+    model::{CheckpointSource, Model},
     model_extra_data::CheckpointData,
 };
 use psyche_core::sha256;
--- a/shared/client/src/state/cooldown.rs
+++ b/shared/client/src/state/cooldown.rs
@@ -1,8 +1,6 @@
 use crate::CheckpointUploader;
 use psyche_coordinator::{
-    CheckpointerSelection, Coordinator,
-    model::{self, LLM, Model},
-    model_extra_data::CheckpointData,
+    CheckpointerSelection, Coordinator, model::Model, model_extra_data::CheckpointData,
 };
 use psyche_data_provider::{GcsManifestMetadata, UploadError, upload_to_gcs, upload_to_hub};
 use psyche_event_sourcing::event;


shared/coordinator/src/coordinator.rs
```diff
@@ -507,6 +513,38 @@ impl<T: NodeIdentity> Coordinator<T> {
...
+        let is_checkpointer = checkpointer_selection
+            .get_checkpointer(witness.proof.index, self.epoch_state.clients.len() as u64);
+        if !is_checkpointer {
+            return Err(CoordinatorError::InvalidWitness);
+        } else {
+            self.epoch_state
+                .checkpointers
+                .push(witness)
+                .map_err(|_| CoordinatorError::WitnessesFull)?;
+        }
```
entropidelic: here we can just run that code without doing an `else` branch

shared/coordinator/src/coordinator.rs
```diff
@@ -602,16 +640,30 @@ impl<T: NodeIdentity> Coordinator<T> {
...
+            return Err(CoordinatorError::InvalidRunState);
+        }
+        let checkpointer_selection = CheckpointerSelection::from_coordinator(self, 0)?;
+        let is_checkpointer = checkpointer_selection
+            .get_checkpointer(index as u64, self.epoch_state.clients.len() as u64);
+        if !is_checkpointer {
+            return Err(CoordinatorError::InvalidWitness);
+        } else {
+            self.epoch_state.checkpointed = true;
+        }
```
entropidelic: here too, we can just run the code without the else branch

shared/client/src/cli.rs
```diff
@@ -231,33 +231,21 @@ impl TrainArgs {
...
-                hub_upload: None,
+                hub_upload: HubUploadInfo {
+                    hub_repo: repo,
+                    hub_token: token.to_string(),
+                },
                 delete_old_steps,
                 keep_steps,
-            }),
-            (_, None, _, _, _) => None,
+            },
```
dsocolobsky: I know you didn't write this function originally but I wonder if we can simplify it with something like this. The match serves no purpose imho.

```suggestion
    if self.keep_steps == 0 {
        bail!("keep_steps must be > 0 for hub repository uploads (got {keep_steps})");
    }

    let hub_token = std::env::var("HF_TOKEN")
        .context("No HF_TOKEN env variable.")?;

    Ok(CheckpointConfig {
        checkpoint_dir: self.checkpoint_dir.clone(),
        hub_upload: HubUploadInfo {
            hub_repo: self.hub_repo.clone(),
            hub_token,
        },
        delete_old_steps: self.delete_old_steps,
        self.keep_steps,
    })
```

IAvecilla: Yeah I changed that function to avoid matching and just check everything manually to exit early, lmk what you think

architectures/centralized/client/src/app.rs
```diff
@@ -174,21 +179,91 @@ impl App {
...
-                    .with_token(Some(hub_token.clone()))
-                    .build()?;
-                let repo_api = api.repo(Repo::new(hub_repo.clone(), hf_hub::RepoType::Model));
-                if !repo_api.is_writable().await {
-                    anyhow::bail!(
-                        "Checkpoint upload repo {} is not writable with the passed API key.",
-                        hub_repo
-                    )
+        let Model::LLM(LLM { checkpoint, .. }) = &self.coordinator_state.model;
+        if !self.skip_upload_check {
```
entropidelic: maybe we can invert the check and return early

shared/client/src/state/cooldown.rs
```diff
@@ -140,97 +146,142 @@ impl CooldownStepMetadata {
...
+                        delete_old_steps,
+                        keep_steps,
+                        hub_token,
+                        skip_upload,
+                    } = checkpoint_info;
+
+                    // When skip_upload is true (testing), skip all checkpoint saving
+                    if skip_upload {
+                        info!("Skipping checkpoint save and upload (skip_upload flag is set)");
+                        checkpoint_completed.store(true, Ordering::SeqCst);
```
entropidelic: you can return `Ok(evals)` early and you don't need to have an `else` branch here

shared/data-provider/src/errors.rs
```diff
@@ -9,45 +10,40 @@ pub enum UploadError {
...
     #[error("GCS authentication failed: {0}")]
-    GcsAuth(#[from] google_cloud_storage::client::google_cloud_auth::error::Error),
+    GcsAuth(String),
 
-    #[error("GCS operation failed: {0}")]
-    GcsStorage(#[from] google_cloud_storage::http::Error),
-
-    // Common errors
+    //#[error("GCS operation failed: {0}")]
+    //GcsStorage(#[from] google_cloud_storage::client::Error),
```
entropidelic: should we delete this?

architectures/centralized/client/src/app.rs
IAvecilla: We have to execute the rest of the code in that function, so I don’t think we can return early there, unless you meant something else?

ethernet8023: considered pulling this out into a fn? this seems like something that could be deduplicated between the centralized and decentralized clients

architectures/centralized/client/src/app.rs
```diff
@@ -173,22 +176,59 @@ impl App {
...
+        // Sanity checks using the checkpoint config from state_options, not the zeroed coordinator state.
+        // The coordinator_state is only populated after receiving the first ServerToClientMessage::Coordinator.
+        if !self.skip_upload_check {
+            let upload_info = match &state_options.checkpoint_config {
+                config if config.skip_upload => Some(UploadInfo::Dummy()),
+                config => {
+                    // Use HF_TOKEN from checkpoint_config for Hub uploads
+                    if let Some(ref hub_token) = config.hub_token {
+                        Some(UploadInfo::Hub(HubUploadInfo {
+                            hub_repo: String::new(), // Will be validated when actual checkpoint is received
```
ethernet8023: this feels a little strange to me, carrying a dummy hub repo around. any way to get this as some sort of enum with a filled/unfilled variant, or something?

architectures/centralized/client/src/main.rs
```diff
@@ -105,7 +105,7 @@ async fn async_main() -> Result<()> {
             )?;
 
             let (mut app, allowlist, p2p, state_options) =
-                build_app(cancel, server_addr, tx_tui_state, args)
+                build_app(cancel, server_addr, tx_tui_state, args, false)
```
ethernet8023: is_testing arg feels a little strange here, just passing a raw boolean.. but maybe this is fine

architectures/decentralized/solana-client/src/app.rs
```diff
@@ -226,6 +232,77 @@ impl App {
         let mut joined_run_this_epoch = None;
         let mut ever_joined_run = false;
 
+        // sanity checks
+        let Model::LLM(LLM { checkpoint, .. }) = start_coordinator_state.model;
+        let upload_info = match checkpoint {
+            model::Checkpoint::Hub(HubRepo { repo_id, revision })
+            | model::Checkpoint::P2P(HubRepo { repo_id, revision }) => {
+                Some(UploadInfo::Hub(HubUploadInfo {
+                    hub_repo: (&repo_id).into(),
```
ethernet8023: weird, does the into() not autoref repo_id?

architectures/decentralized/solana-client/src/app.rs
```diff
@@ -226,6 +232,77 @@ impl App {
...
+                let client = StorageControl::builder().build().await?;
+
+                let permissions_to_test = vec![
+                    "storage.objects.list",
+                    "storage.objects.get",
+                    "storage.objects.create",
+                    "storage.objects.delete",
+                ];
+
+                let resource = format!("projects/_/buckets/{}", gcs_info.gcs_bucket);
```
ethernet8023: this seems like a lot of GCS-specific stuff to have in the main app file. have we considered making a tiny wrapper module in some other crate that makes this api a bit more like the hf one? or even a trait to abstract reading / writing checkpoints?

shared/client/src/state/cooldown.rs
```diff
@@ -140,59 +146,104 @@ impl CooldownStepMetadata {
...
-                    delete_old_steps,
-                    keep_steps,
-                }) = checkpoint_info
-                else {
-                    return Ok((evals, None));
-                };
-
-                let upload_handle = tokio::task::spawn(async move {
+
+                    let upload_info = match checkpoint {
```
ethernet8023: this kind of duplication of structure makes me wonder if we shouldn't have a separate enum for, like
```rs
enum HostedModel {
  Gcp { whatever }
  Hub { whatever }
}
```

and then a checkpoint would be 
```rs
enum Checkpoint {
  Hosted(HostedModel)
  P2P(HostedModel)
}
```

just to remove this weird duplication of p2p / p2pgcs. thoughts?

shared/client/src/cli.rs
```diff
@@ -329,6 +269,13 @@ impl TrainArgs {
     }
 }
 
+fn default_checkpoint_dir() -> PathBuf {
```
ethernet8023, line 312: should this not write to somewhere in the scratch dir in the container, so all our files end up in the same place?

shared/coordinator/src/checkpointer_selection.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    pub fn from_coordinator<T: NodeIdentity>(
+        coordinator: &Coordinator<T>,
+        offset: isize,
+    ) -> Result<Self, CoordinatorError> {
+        let round = get_round_by_offset(coordinator, offset)?;
+        let seed = sha256(&round.random_seed.to_le_bytes());
+
+        let checkpointers = max(
+            (coordinator.epoch_state.clients.len() / 3).min(SOLANA_MAX_NUM_CHECKPOINTERS),
```
ethernet8023, line 30: will this be correct if a client has requested to be withdrawn e.g. within warmup? (it might be, just sanity checking)

shared/coordinator/src/model.rs
```diff
@@ -280,10 +293,10 @@ impl GcsRepo {
 #[repr(C)]
 pub enum Checkpoint {
     Ephemeral,
-    Dummy(HubRepo),
+    Dummy(HubRepo), // Used for testing
     Hub(HubRepo),
-    P2P(HubRepo),
     Gcs(GcsRepo),
+    P2P(HubRepo),
```
ethernet8023: doesn't re-ordering here mean that existing coordinators currently set to P2P will suddenly decode as Gcs?

tools/rust-tools/run-manager/src/docker/manager.rs
```diff
@@ -163,8 +163,17 @@ impl RunManager {
             .arg(&self.env_file);
 
         if let Some(dir) = &self.scratch_dir {
```
ethernet8023, line 224: doesn't this assert that you MUST pass google credentials for any run started via run-manager, meaning we can't do e.g. hf checkpointing anymore?

also, i would much rather we find a way to pass the creds via some other mechanism - maybe write them to a temp dir the same way we write the env file, or something? adding another "you must have this file in this dir" thing is janky IMO

architectures/centralized/client/src/app.rs
IAvecilla: Totally true. Since right now the only way to get the checkpoint is through the Coordinator, we have to update the source once the client connects and receives the updates. I refactored it to use an enum that carries the minimum required info at first, and then switches to the other variant once all the necessary information is available.

architectures/centralized/client/src/main.rs
IAvecilla: Yeah, I don’t like it either, but it was a quick way to avoid checking credentials and permissions for uploading to external storage. I tried another approach to avoid adding new flags everywhere.

architectures/decentralized/solana-client/src/app.rs
IAvecilla: With some refactors I made, I think this part is no longer needed. But yeah, I thought the same when using `into`, and it was kind of suggesting that approach.

architectures/decentralized/solana-client/src/app.rs
IAvecilla: I moved all the credential-specific logic into a new module, where I defined the `UploadInfo` and `UploadCredentials` structs, and we only call `validate_credentials` from there. Maybe the trait approach is even better, so that’s a good suggestion, I’ll try that. But at least this should be an improvement.

shared/client/src/state/cooldown.rs
IAvecilla: The only reason I did this is that adding a new enum containing those variants would change the size of the `Checkpoint` struct, which would in turn change the Coordinator’s memory layout. I wanted to avoid a redeployment caused by storage changes. I much prefer the suggested approach, but to avoid dealing with those changes I went in this direction, does that make sense?

shared/client/src/cli.rs
IAvecilla: Yep, good catch. I changed it to check whether we have the scratch directory and to keep all the local checkpoints there.

shared/coordinator/src/checkpointer_selection.rs
IAvecilla: I think that on every state transition we call `move_clients_to_exited` to check whether any clients are unhealthy and move them from `epoch_state.clients` to `epoch_state.exited_clients`, so that should handle it. Worst case, if we don’t move one, the idea is to select multiple checkpointers at the end of an epoch to increase the chances that at least one of them is healthy and successfully uploads the full model.

shared/coordinator/src/model.rs
IAvecilla: YES! nice catch, I changed it to keep the original order 🙂

tools/rust-tools/run-manager/src/docker/manager.rs
IAvecilla: Yes! Also, thank you for catching that. I changed it to be optional and now we only mount the credentials into Docker if they exist locally, but this should support Hugging Face the same way as before.

I agree that asking to move the credentials to the scratch dir is kind of annoying. The only alternative I can think of is setting `GOOGLE_CREDENTIALS_FILE_PATH` in the `.env` to point to the local credential file, and then having the manager script automatically mount that file into the scratch dir and set the env variable. Does that sound better?

ethernet8023: > I agree that asking to move the credentials to the scratch dir is kind of annoying. The only alternative I can think of is setting GOOGLE_CREDENTIALS_FILE_PATH in the .env to point to the local credential file, and then having the manager script automatically mount that file into the scratch dir and set the env variable. Does that sound better?

Yeah, i think this would work better if that's cool :)

.github/workflows/solana-build-anchor-programs.yml
```diff
@@ -19,7 +19,7 @@ jobs:
...
-          key: validator-image-${{ runner.os }}-${{ hashFiles('shared/coordinator/src/coordinator.rs', 'architectures/decentralized/solana-coordinator/**/*.rs', 'architectures/decentralized/solana-coordinator/**/*.toml', 'architectures/decentralized/solana-coordinator/Cargo.lock', 'architectures/decentralized/solana-authorizer/**/*.rs', 'architectures/decentralized/solana-authorizer/**/*.toml', 'architectures/decentralized/solana-authorizer/Cargo.lock', 'docker/test/psyche_solana_validator_entrypoint.sh', 'nix/docker.nix', 'flake.lock') }}
+          key: validator-image-${{ runner.os }}-${{ hashFiles('shared/coordinator/src/coordinator.rs', 'shared/coordinator/src/model.rs', 'architectures/decentralized/solana-coordinator/**/*.rs', 'architectures/decentralized/solana-coordinator/**/*.toml', 'architectures/decentralized/solana-coordinator/Cargo.lock', 'architectures/decentralized/solana-authorizer/**/*.rs', 'architectures/decentralized/solana-authorizer/**/*.toml', 'architectures/decentralized/solana-authorizer/Cargo.lock', 'docker/test/psyche_solana_validator_entrypoint.sh', 'nix/docker.nix', 'flake.lock') }}
```
ethernet8023, line 22: noting that the comment is true and we can keep adding files to this but eventually w should fix it for real 😅

architectures/decentralized/solana-client/src/app.rs
```diff
@@ -230,6 +234,30 @@ impl App {
...
+                    .checkpoint_config
+                    .hub_token
+                    .as_ref()
+                    .map(|token| UploadCredentials::HubToken(token.clone())),
+                model::Checkpoint::Gcs(GcsRepo { bucket, .. })
+                | model::Checkpoint::P2PGcs(model::GcsRepo { bucket, .. }) => {
+                    Some(UploadCredentials::GcsBucket(bucket.to_string()))
+                }
+                _ => None,
+            };
```
ethernet8023: So, if the model's init checkpoint is e.g. HuggingFace at a specific repo, all subsequent uploads must be to that *same repo*?

Noting that this is a loss of functionality from the previous behavior, where each individual client could have its own repo / upload location for checkpoints.

Not sure if this is an important feature to keep, but worth considering

shared/client/src/state/types.rs
```diff
@@ -9,18 +10,124 @@ use tch::TchError;
...
+                .all(|p| response.permissions.contains(&p.to_string()));
+            if !correct_permissions {
+                anyhow::bail!(
+                    "GCS bucket {} does not have the required permissions for checkpoint upload. Make sure to set GOOGLE_APPLICATION_CREDENTIALS environment variable correctly and have the correct permissions to the bucket.",
+                    bucket
+                )
+            }
+        }
+        Ok(())
+    }
```
ethernet8023: any reason this is a separate method than the validate() method, and not just part of calling .validate()?

shared/client/src/state/types.rs
```diff
@@ -9,18 +10,124 @@ use tch::TchError;
...
+    /// Skip validation (for testing)
+    Skip,
+}
+
+impl UploadCredentials {
+    /// Validates that the upload credentials are valid.
+    pub async fn validate(&self) -> anyhow::Result<()> {
+        match self {
+            UploadCredentials::HubToken(token) => {
+                let _api = hf_hub::api::tokio::ApiBuilder::new()
```
ethernet8023: a regression: we no longer validate that the specific repo we want to upload to is writable by our hf hub token

architectures/centralized/client/src/app.rs
```diff
@@ -167,22 +167,29 @@ impl App {
...
+        // Sanity checks using the checkpoint config from state_options, not the zeroed coordinator state.
+        // The coordinator_state is only populated after receiving the first ServerToClientMessage::Coordinator.
+        if !state_options.checkpoint_config.skip_upload {
+            let credentials = if let Some(ref hub_token) = state_options.checkpoint_config.hub_token
             {
-                let api = hf_hub::api::tokio::ApiBuilder::new()
-                    .with_token(Some(hub_token.clone()))
-                    .build()?;
-                let repo_api = api.repo(Repo::new(hub_repo.clone(), hf_hub::RepoType::Model));
-                if !repo_api.is_writable().await {
```
ethernet8023, line 186: we seem to have lost this check that our HF token has the ability to write to the repo in question.

architectures/centralized/client/src/app.rs
```diff
@@ -167,22 +167,29 @@ impl App {
...
+                None => {
                     anyhow::bail!(
-                        "Checkpoint upload repo {} is not writable with the passed API key.",
-                        hub_repo
-                    )
+                        "No upload credentials found for checkpointing. Set HF_TOKEN for HuggingFace Hub or configure GCS credentials."
+                    );
                 }
             }
         }
```
ethernet8023: if `The coordinator_state is only populated after receiving the first ServerToClientMessage::Coordinator.`, 
we could instead wait for the first state after ::Join, then run this validation on that state, then pass it to `rx_from_server_message`, then construct the client?

architectures/centralized/client/src/main.rs
```diff
@@ -129,6 +129,10 @@ async fn async_main() -> Result<()> {
 }
 
 fn main() -> Result<()> {
+    rustls::crypto::ring::default_provider()
+        .install_default()
+        .expect("Failed to install default CryptoProvider");
+
```
ethernet8023, line 153: maybe we should build some common init stuff... worried about remembering this everywhere lol

architectures/centralized/testing/src/test_utils.rs
```diff
@@ -127,6 +127,7 @@ pub fn dummy_client_app_params_with_training_delay(
     run_id: &str,
     training_delay_secs: u64,
 ) -> AppParams {
+    std::env::set_var("HF_TOKEN", "dummy_token");
```
ethernet8023, line 130: a thought that we should consider refactoring all our secrets stuff - wallet keymaterial, hf token, et al - in a future PR.

architectures/decentralized/solana-client/src/app.rs
```diff
@@ -230,6 +234,30 @@ impl App {
...
+            if let Some(ref creds) = credentials {
+                // Validate basic credentials
+                creds.validate().await?;
+                // For GCS, also validate bucket permissions
+                creds.validate_gcs_bucket_permissions().await?;
+            }
```
ethernet8023: the idea of creating creds and validating them in two steps is a little smelly to me -
it means we can instantiate a credentials type that's not valid
i'd rather some `CheckpointUploader` struct that abstracts away the upload code itself for each provider, and can only be constructed via some async method that actually validates the credentials on initialization. maybe we remove the `UploadCredentials` struct and thread the uploader w/ all info needed to upload all the way through?

this does come with the weird case of needing to construct a new one for if the run gets paused & target upload bucket/hf repo changed, but this existing credential check would miss verifying that anyways.

```rs
let checkpoint_uploader = CheckpointUploader::new_hf(repo, token).await?
/// or.
let checkpoint_uploader = CheckpointUploader::new_gcs(repo).await?
```

would fail if the creds are not valid for the repo in question.

we could then thread this uploader through wherever needed.
maybe we should create this whenever the checkpoint changes, just to verify we still have access...

architectures/decentralized/solana-client/Cargo.toml
```diff
@@ -29,8 +29,11 @@ time.workspace = true
 tokio.workspace = true
 tokio-util.workspace = true
 tracing.workspace = true
+google-cloud-storage.workspace = true
+hf-hub.workspace = true
 psyche-python-extension-impl = { workspace = true, optional = true }
 tikv-jemallocator.workspace = true
+rustls = { version = "0.23", default-features = false, features = ["ring"] }
```
ethernet8023: we should put this in the workspace's toml and do `rustls.workspace = true` for this and centralizxed

architectures/decentralized/justfile
```diff
@@ -37,10 +39,16 @@ setup-solana-localnet-permissioned-light-test-run-treasurer run_id="test" *args=
...
 
 start-training-localnet-light-client run_id="test" *args='':
-    AUTHORIZER={{ AUTHORIZER }} RUN_ID={{ run_id }} BATCH_SIZE=1 DP=1 ./scripts/train-solana-test.sh {{ args }}
+    HF_TOKEN={{ HF_TOKEN }} GOOGLE_APPLICATION_CREDENTIALS={{ GOOGLE_APPLICATION_CREDENTIALS }} AUTHORIZER={{ AUTHORIZER }} CHECKPOINT="false" RUN_ID={{ run_id }} BATCH_SIZE=1 DP=1 ./scripts/train-solana-test.sh {{ args }}
+
+start-training-localnet-light-client-checkpoint run_id="test" *args='':
+    HF_TOKEN={{ HF_TOKEN }} GOOGLE_APPLICATION_CREDENTIALS={{ GOOGLE_APPLICATION_CREDENTIALS }} AUTHORIZER={{ AUTHORIZER }} CHECKPOINT="true" RUN_ID={{ run_id }} BATCH_SIZE=1 DP=1 ./scripts/train-solana-test.sh {{ args }}
+
+start-training-localnet-client-checkpoint run_id="test" *args='':
+    HF_TOKEN={{ HF_TOKEN }} GOOGLE_APPLICATION_CREDENTIALS={{ GOOGLE_APPLICATION_CREDENTIALS }} AUTHORIZER={{ AUTHORIZER }} CHECKPOINT="true" RUN_ID={{ run_id }} ./scripts/train-solana-test.sh {{ args }}
```
ethernet8023, line 51: we're starting to get lots of env vars here. maybe split these justfiles into some modules and use "export" to bring the vars in automatically?

scripts/train-solana-test.sh
```diff
@@ -20,23 +20,31 @@ elif [[ -z "${WALLET_FILE:-}" ]]; then
...
+else
+    echo -e "\n[+] Starting Solana training without checkpointing..."
+fi
+
 # presets for a DGX or an HGX
 DP=${DP:-"8"}
 TP=${TP:-"1"}
 BATCH_SIZE=${BATCH_SIZE:-"1"}
+HF_TOKEN=${HF_TOKEN:-""}
+GOOGLE_APPLICATION_CREDENTIALS=${GOOGLE_APPLICATION_CREDENTIALS:-""}
```
ethernet8023: you can just `export` these instead of writing the name twice to pass em. unimportant though

shared/client/src/state/cooldown.rs
```diff
@@ -139,59 +146,104 @@ impl CooldownStepMetadata {
...
+                                .await
+                                .map_err(|_| CheckpointError::ExtractThreadCrashed)?
+                        }
+                        _ => (variables, trainer),
+                    };
+
+                    trainers.push(trainer);
+                    let evals = model_task_runner.start(trainers);
+                    if !is_checkpointer {
+                        info!("Skipping checkpoint upload as this node is not the checkpointer for this epoch");
```
ethernet8023: "a" checkpointer instead perhaps

shared/client/src/state/cooldown.rs
ethernet8023: we have to change stoage anyways now, so let's get it 😁

shared/client/src/state/cooldown.rs
```diff
@@ -221,16 +275,15 @@ impl CooldownStepMetadata {
...
                     .await;
 
-                    Ok(())
-                });
-
-                Ok((evals, Some(upload_handle)))
-            }
-            .instrument(info_span!("checkpointing")),
-        );
+                    Ok(evals)
```
ethernet8023, line 289: all these if/else/checkpoint_completed things are a little suspicious to me. i wonder  if we can do something with a try block, so we can store checkpoint completed in one place if the try block exits with Ok, and then do the cleanup_dirs regardless, and keep the error! thing only if there's an errorr returned from the try block?

shared/client/src/state/cooldown.rs
```diff
@@ -264,50 +317,50 @@ async fn upload_checkpoint(
...
-            upload_to_gcs(gcs_info, manifest_metadata, local, step, tx_checkpoint)
+            upload_to_gcs(gcs_info, manifest_metadata, local, step, cancellation_token)
                 .await
                 .map_err(CheckpointError::UploadError)
         }
-        UploadInfo::Hub(hub_info) => upload_to_hub(hub_info, local, step, tx_checkpoint)
+        UploadInfo::Hub(hub_info) => upload_to_hub(hub_info, local, step, cancellation_token)
             .await
             .map_err(CheckpointError::UploadError),
+        UploadInfo::Dummy() => {
```
ethernet8023:
```suggestion
        UploadInfo::Dummy => {
```

shared/client/src/state/types.rs
```diff
@@ -9,18 +10,124 @@ use tch::TchError;
 use thiserror::Error;
 use tokio::task::JoinHandle;
 
+/// Upload info with full details needed for actual uploads.
 #[derive(Debug, Clone)]
 pub enum UploadInfo {
     Hub(HubUploadInfo),
     Gcs(GcsUploadInfo),
+    Dummy(),
```
ethernet8023:
```suggestion
    Dummy,
```

shared/client/src/state/init.rs
```diff
@@ -457,7 +455,7 @@ impl RunInitConfigAndIO {
                                     .collect();
                                 let tokenizer = Arc::new(auto_tokenizer(&repo_files)?);
                                 (
-                                    PretrainedSource::<AutoConfig>::RepoFiles(repo_files),
+                                    PretrainedSource::<AutoConfig>::RepoFiles(repo_files.to_vec()),
```
ethernet8023:
```suggestion
                                    PretrainedSource::<AutoConfig>::RepoFiles(repo_files),
```
it's already a vec

shared/client/src/state/steps.rs
```diff
@@ -163,180 +161,235 @@ impl StepStateMachine {
...
-                let all_prev_round_batches_are_trained = self
-                    .previous_round
-                    .batch_ids_not_yet_trained_on
-                    .lock()
-                    .unwrap()
-                    .is_none();
-
-                if step.finished() && all_prev_round_batches_are_trained {
-                    // Finished training and finished downloading the previous round's results
-                    // (or we're on the first or last which has nothing to download)
```
ethernet8023, line 185: any reason to remove this part?

shared/client/src/state/steps.rs
```diff
@@ -163,180 +161,235 @@ impl StepStateMachine {
...
+    }
+
+    fn try_send_cooldown_witness(&mut self) -> Result<(), OpportunisticWitnessError> {
+        if let ActiveStep::Cooldown(active_step) = &self.active_step {
+            if !active_step.checkpoint_complete() || self.sent_cooldown_witness {
+                return Ok(());
             }
         }
+
+        // Original behavior: panic if client not found (should never happen)
```
ethernet8023: "original behavior" i think we can remove these words

shared/client/src/state/steps.rs
```diff
@@ -163,180 +161,235 @@ impl StepStateMachine {
             step_finish_time: None,
             sent_warmup_finished: false,
             sent_warmup_witness: false,
-
-            pending_upload_handles: Vec::new(),
+            sent_cooldown_witness: false,
```
ethernet8023: i wonder if we can move these into their respective steps, rather than keeping them in the main stepstatemachine

shared/client/src/state/steps.rs
```diff
@@ -834,19 +884,19 @@ impl StepStateMachine {
...
             (ActiveStep::Cooldown(cooldown), RunState::WaitingForMembers)
             | (ActiveStep::Cooldown(cooldown), RunState::Warmup)
             | (ActiveStep::Cooldown(cooldown), RunState::Paused) => {
-                let (trainers, upload_handle) = cooldown.finish().await?;
-                if let Some(handle) = upload_handle {
-                    self.pending_upload_handles.push(handle);
-                }
+                // If we reach state it means at least one of the clients has successfully uploaded the model checkpoint.
+                // We can cancel any of the other uploads in progress.
+                cooldown.cancel();
```
ethernet8023: any reason this can't be part of `.finish()`?

shared/client/src/state/types.rs
```diff
@@ -9,18 +10,124 @@ use tch::TchError;
...
+}
+
+impl UploadCredentials {
+    /// Validates that the upload credentials are valid.
+    pub async fn validate(&self) -> anyhow::Result<()> {
+        match self {
+            UploadCredentials::HubToken(token) => {
+                let _api = hf_hub::api::tokio::ApiBuilder::new()
+                    .with_token(Some(token.clone()))
+                    .build()?;
```
ethernet8023: this doesn't actually validate anything anymore, this just will work no matter what

shared/client/src/cli.rs
```diff
@@ -329,6 +269,17 @@ impl TrainArgs {
...
+    let final_dir = if std::path::Path::new("/scratch").exists() {
```
ethernet8023, line 313: the more /scratch work we do the more i wonder if we should just have a `PSYCHE_WORKING_DIR` or something rather than trying so many paths

shared/coordinator/src/committee_selection.rs
```diff
@@ -1,424 +0,0 @@
-use crate::{Client, Coordinator, CoordinatorError, SOLANA_MAX_NUM_WITNESSES};
```
ethernet8023, line 1: why split committe_selection into committee.rs and tests.rs?

shared/coordinator/src/model.rs
```diff
@@ -239,6 +241,17 @@ impl HubRepo {
     }
 }
 
+impl FromStr for HubRepo {
+    type Err = anyhow::Error;
```
ethernet8023: I'd like to keep anyhow out of the coordinator crate if possible - looks like this is impossible to fail, too? should we just impl `From<&str>`?

.github/workflows/solana-build-anchor-programs.yml
IAvecilla: Agree, I'll open a PR for a real follow-up on this

architectures/decentralized/solana-client/src/app.rs
IAvecilla: Yeah exactly, the idea behind all of this was that the run owner has one place to get the model output and gets total control of where the model checkpoints (and final output) reside. So all the clients should only checkpoint to the original model source. I'm open to discuss it further tho

shared/client/src/state/types.rs
IAvecilla: Not really, I think it's safe to merge the code in one same function

shared/client/src/state/types.rs
IAvecilla: Yeah, that's bad. I added `CheckpointUploader::new_hub()` (following later suggestions) and now calls `is_writable()` to verify write access to the specific repo, matching the GCS bucket permission check. This was actually removed on main before this PR, so we're restoring it.

architectures/centralized/client/src/app.rs
IAvecilla: Totally, I restructured the centralized client to wait for the first coordinator state after joining, then validate credentials with the actual repo info before proceeding.

architectures/centralized/client/src/app.rs
IAvecilla: It's a very good idea so I followed that exact implementation

architectures/centralized/client/src/main.rs
IAvecilla: Yeah that would be great, I'll create an issue to address this in the future to avoid adding more code to this PR but It could be a good refactor

architectures/centralized/testing/src/test_utils.rs
IAvecilla: Also good idea, let me know if you have an approach in mind and we can create an issue explaining it and see the best way to implement it

architectures/decentralized/solana-client/src/app.rs
IAvecilla: I ended up replacing both `UploadCredentials` and `UploadInfo` with `CheckpointUploader` which validates credentials and write permissions on construction (new_hub/new_gcs). It's constructed at startup for fail-fast validation, and re-created during each cooldown from the current coordinator state so permissions are re-verified if the checkpoint target changes mid-run.

architectures/decentralized/justfile
IAvecilla: Good idea, will tackle as a follow-up

shared/client/src/state/cooldown.rs
IAvecilla: I consolidated `checkpoint_completed` to be set in a single place instead of three. Considered the try block approach but the "no uploader" cases (missing HF token, Ephemeral/Dummy checkpoint) are valid success paths, not errors, so they'd need special `return Ok(())` handling inside the block. The current flow handles those naturally and `checkpoint_completed` is already set in one place but let me know if I didn't understand something correctly with the approach

shared/client/src/state/steps.rs
IAvecilla: Yeah I removed `doing_checkpoint()` / `pending_upload_handles` because uploads now run inline in the cooldown task instead of as fire-and-forget background handles. However, It's true that we might be losing some functionality there not checkpointing correctly when pausing so I fixed the pause behavior. On Paused, we now wait for the upload to finish if there's some and I made sure to not crash the exit the client until this completes

shared/client/src/state/steps.rs
IAvecilla: Looked into this, the methods that use these booleans depend heavily on `StepStateMachine` fields (coordinator state, channels, identity, etc.). Moving them would require threading all those dependencies into the step structs. Maybe we can check an alternative in a future PR, I agree that it will be cleaner and it makes more sense but didn't wanna do a whole refactor of this to avoid adding extra logic here

IAvecilla: Okay nevermind I ended up implementing it because it wasn't that much code and I think it's way better than before

shared/client/src/state/steps.rs
IAvecilla: Not really, it's fixed

shared/client/src/state/types.rs
IAvecilla: I changed a lot of the validation on the tokens and the permissions so it should make sense now

shared/client/src/cli.rs
IAvecilla: That's not a bad idea. I think it would also be more descriptive than using scratch as a name, so it's a good follow-up.

shared/coordinator/src/committee_selection.rs
IAvecilla: At first my idea was to include tests for both `CommitteeSelection` and `CheckpointerSelection` in a single file, but I realized the tests for `CheckpointerSelection` were essentially the same as the ones for the `Committee`, so I reverted that change. Now the tests and the committee logic are in the same file. I also renamed the new file from `checkpointer.rs` to `checkpointer_selection.rs` for consistency.

PR #457, opened 2026-01-02
feat: implement authorization management directly in CLI

crypto-vincent:
## Summary

Port authorization management scripts to the psyche-client CLI's rust implementation.

One small detail also:
> The delegate-clear flag is now default to false when setting deletes on an authoriztion

### DIFF
--- a/architectures/decentralized/solana-client/src/command/can_join.rs
+++ b/architectures/decentralized/solana-client/src/command/can_join.rs
@@ -49,7 +49,10 @@ pub async fn command_can_join_execute(
         authorizer,
     );
     if backend.get_balance(&authorization).await? == 0 {
-        bail!("Authorization does not exist for authorizer: {authorizer:?} and user: {address}");
+        bail!(
+            "Authorization does not exist for authorizer: {authorizer:?} (authorization address: {authorization:?}, join authority: {0:?}). Authorizer must be set to grantee pubkey for permissioned runs",
+            coordinator_instance_state.join_authority
+        );
     }
     if !backend
         .get_authorization(&authorization)


psyche-book/src/enduser/authentication.md
```diff
@@ -12,50 +12,64 @@ The following roles will be important:
...
+A run can be made permissionless by creating an `authorization` with a special `authorizer` valid for everyone: `11111111111111111111111111111111`
+
+A CLI is provided for this:
+
+```sh
+psyche-solana-client join-authorization-create \
+    --rpc [RPC] \
+    --wallet-private-key-path [JOIN_AUTHORITY_KEYPAIR_FILE] \
+    --authorizer 11111111111111111111111111111111
+```
```
entropidelic, line 32: instead of having to set `11111111111111111111111111111111` manually, I think it could be useful to define a new flag (maybe `--permissionless`) that conflicts with the `--authorizer` argument so that only one of the two is set when the command runs. The intent seems much clearer this way

crypto-vincent: That's a good point. I'd like to do that in a separate PR because if we do apply this logic, we'd need to modify the following commands for homogeneity:

- create-run
- join-run (maybe)
- join-authorization-[create/delete/read]

So that we use the same vocabulary in all places to avoid confusion. As of right now, all commands have the "authorizer" set to 1111...1111 explicitely, both during reading and writing so while it might be less intuitive its more cohesive in the current state IMO. Definitly willing to clarify/simplify this API later if this is too confusing.

We could make an issue to unify/simplify this specifically ? where would be a good place for tracking this ?

entropidelic (approved):

PR #479, opened 2026-01-12
broadcast parameter names along with config on p2p network

jquesnelle:


shared/network/src/p2p_model_sharing.rs
```diff
@@ -473,8 +480,23 @@ impl SharableModel {
...
+                        .keys()
+                        .cloned()
+                        .chain(
+                            self.serializing_parameters
+                                .unwrap_or_default()
+                                .keys()
+                                .cloned(),
+                        )
+                        .collect::<HashSet<String>>()
+                        .iter()
```
ethernet8023: should be `.into_iter()`.

shared/network/src/p2p_model_sharing.rs
```diff
@@ -473,8 +480,23 @@ impl SharableModel {
...
+                        .unwrap_or_default()
+                        .chain(
+                            self.serializing_parameters
+                                .as_ref()
+                                .map(|x| x.keys().cloned())
+                                .unwrap_or_default(),
+                        )
+                        .collect::<HashSet<String>>()
+                        .into_iter()
+                        .collect::<Vec<String>>(),
```
IAvecilla: I think we can use the following code to get the parameter names
```
 self
      .parameters
      .as_ref()
      .ok_or_else(|| SharableModelError::ModelConfigNotInitialized)?
      .keys()
      .cloned()
      .collect::<Vec<_>>();
```

shared/modeling/src/models/deepseek.rs
```diff
@@ -1036,97 +1036,7 @@ impl DeepseekForCausalLM {
 
 impl ModelConfig for DeepseekConfig {
     fn get_parameter_names(&self) -> Vec<String> {
```
IAvecilla: I think we can remove the `ModelConfig` trait everywhere since this is the only function it has and we're not using it anymore

jquesnelle: I tried and it turned into a giant mess but if you can do it would be much appreciated!

shared/network/src/p2p_model_sharing.rs
jquesnelle: yeah this is better, thanks

Issue #469, opened 2026-01-08
Migrate run administration tools into the run manager


ethernet8023:
[https://github.com/PsycheFoundation/psyche/blob/main/psyche-book/src/enduser/authentication.md](https://github.com/PsycheFoundation/psyche/blob/main/psyche-book/src/enduser/authentication.md)<br>These commands, and commands to create and manage runs, are currently built into the training client. We should move them into the run manager.

PR #480, opened 2026-01-12
move admin comands from solana-client to run-manager

ethernet8023:
run-manager binary - moved all admin commands here:
- run management: create-run, close-run, update-config, set-paused, set-future-epoch-rates, tick, checkpoint, json-dump-run, json-dump-user
- authorization: join-authorization-{create,read,delegate,delete}
- treasurer: treasurer-{claim-rewards,top-up-rewards}
- validation: can-join

add run-manager tests
better nix build system for all rust binaries, less jank and automatic binary discovery
added all cli commands to docs (autogenerated !)
factor out some common solana stuff into a new crate

breaking changes: `predownload` is now a subcommand in psyche-solana-client - rather than doing `psyche-solana-client can-join --predownload-model`, it's simply `psyche-solana-client predownload`. `can-join` is now in the run-manager.

closes https://github.com/PsycheFoundation/psyche/issues/469

### DIFF
--- a/.github/workflows/build-run-manager.yml
+++ b/.github/workflows/build-run-manager.yml
@@ -2,10 +2,17 @@ name: Build run-manager
 
 on:
   push:
-    paths:
-      - 'tools/rust-tools/run-manager/**'
-      - '.github/workflows/build-run-manager.yml'
+    branches:
+      - main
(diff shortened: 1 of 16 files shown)


jquesnelle:
@arilotter the run manager version pr conflicted it -- can you fix it up?

PR #485, opened 2026-01-13
GCS Model: manifest.json

pefontana:
Adds manifest.json for GCS downloads. 
The manifest.json is a metadata file uploaded alongside safetensors that tracks file generations for cache invalidation.
Also, implement some changes in the psyche cache, to ensure we are downloading the actual model

  Changes:
  -  Two download paths: download_files_from_manifest for checkpoints with manifest, download_files_no_manifest as fallback for buckets that dont have a manifest.json
  - Upload manifest.json alongside checkpoint files with metadata (step, epoch, run_id) and file generations
  - Version cache folders using step and manifest generation: step-{step}-{generation}/
  - Download config files (.json, .py) separately after manifest download

### DIFF
--- a/shared/data-provider/src/gcs.rs
+++ b/shared/data-provider/src/gcs.rs
@@ -271,8 +271,9 @@ async fn download_files_no_manifest(
     }
 
     info!(
-        "Found {} model files in gs://{}/{}",
+        "Found {} files ({}) in gs://{}/{}",
         all_objects.len(),
+        extensions.join(", "),
         bucket,
         prefix.unwrap_or("")
     );
@@ -289,7 +290,7 @@ async fn download_files_no_manifest(
             continue;
         }
 
-        info!("Downloading: {}", object_name);
+        info!("Downloading: gs://{}/{}", bucket, object_name);
 
         let data = client
             .download_object(


shared/data-provider/src/gcs.rs
```diff
@@ -8,24 +9,77 @@ use google_cloud_storage::http::objects::{
...
+}
+
 const MODEL_EXTENSIONS: [&str; 3] = [".safetensors", ".json", ".py"];
 
 fn check_model_extension(filename: &str) -> bool {
     MODEL_EXTENSIONS.iter().any(|ext| filename.ends_with(ext))
 }
 
-fn get_cache_dir(bucket: &str, prefix: Option<&str>) -> PathBuf {
+fn get_cache_dir(
```
jquesnelle, line 70: we should figure out how this plays with `SCRATCH_DIR`, `CHECKPOINT_DIR`, and `HF_HOME`.

we currently mount `SCRATCH_DIR` as an external docker mount, giving persistent access through docker reboots (maybe we need to rename this to `PERSISTENT_DIR`?)

then, we use `HF_HOME` for hf caching -- typically we'd set `HF_HOME` to the same value as `SCRATCH_DIR`, and then `CHECKPOINT_DIR` is used for saving checkpoints. to make this work as closely as it does through the HF_HOME flow, perhaps we use `HF_HOME` here? (and store everything in a `psyche/gcs` subfolder

pefontana: Sure!
We can use `HF_HOME` for now and then when we remove HuggingFace checkpoint rename it

PR #506, opened 2026-01-16
HF-preprocessed provider

dsocolobsky:
Adds a new `HuggingFacePreprocessedDataProvider` Data Provider which fetches pre-tokenized datasets from HuggingFace via their streaming interface (so, on-demand).

I added a new test config `hf-preprocessed-config.toml` using `emozilla/Hermes-3-Preprocessed-Llama3`, training seemed to work alright but I'm not that familiar with these datasets/api so requires further testing.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7719,7 +7719,7 @@ dependencies = [
 [[package]]
 name = "pyo3-tch"
 version = "0.22.0"
-source = "git+https://github.com/jquesnelle/tch-rs.git?rev=dda507e05a776547a112b6854d1e611684f8c729#dda507e05a776547a112b6854d1e611684f8c729"
+source = "git+https://github.com/NousResearch/tch-rs.git?rev=dda507e05a776547a112b6854d1e611684f8c729#dda507e05a776547a112b6854d1e611684f8c729"
 dependencies = [
  "pyo3",
  "tch",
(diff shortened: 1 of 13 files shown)


shared/data-provider/src/hf_preprocessed.rs
```diff
@@ -0,0 +1,314 @@
...
+            bail!("No data available");
+        }
+
+        // Batch consecutive indices together for efficient API calls
+        let row_indices = self.map_batch_id_to_rows(data_ids);
+        let mut samples = Vec::with_capacity(row_indices.len());
+
+        // For example if row_indices = [5, 6, 7, 10, 11] then the loop does:
+        // - [5, 6, 7] in first iteration since they are consecutive
+        // - [10, 11] in second iteration
```
IAvecilla, line 228: I wonder how this works with the shuffled indices. There’s a small chance we could end up with consecutive numbers, which would mean making one request per index, so we’d likely get rate-limited sooner rather than later. Maybe we could request the data using ordered indices and then shuffle the results afterward.

shared/data-provider/src/hf_preprocessed.rs
```diff
@@ -0,0 +1,314 @@
...
+            labels,
+            position_ids,
+            sequence_lengths,
+        })
+    }
+
+    fn map_batch_id_to_rows(&self, batch_id: BatchId) -> Vec<usize> {
+        if self.num_rows == 0 {
+            return vec![];
+        }
```
IAvecilla, line 204: I think we're already checking this in the `get_samples` function before calling this, can this be removed?

dsocolobsky: I think it's worth it to leave the check just in case at some point we re-use it somewhere. If we ever call this before checking before if num_rows=0 it will access out of bounds and panic. I prefer not to have this hidden condition.

shared/data-provider/src/hf_preprocessed.rs
dsocolobsky: I've taken a look at this but at a first glance it seems like we can't avoid fetching 1 by 1 in the case the indices are shuffled. It's very unlikely that if we have e.g. 4 shuffled indices in the range [0,40000] some of them will be consecutive.

PR #511, opened 2026-01-22
Adds HTTP to train.rs example

philrhc:
Merges the HTTP example with the train example


philrhc:
Hi @IAvecilla, i saw this comment: https://github.com/PsycheFoundation/psyche/pull/506#issuecomment-3780359474
It would make it clear to me if you could give me an example URL or expand on your comment in the context of this file. I've tried the HTTP template example with these parameters and it seems to work, I suppose this not pre-processed data from huggingface?
```
cargo run --example train --     --model emozilla/llama2-20m-init     --total-batch 2     --micro-batch 1     http-template         --template "https://huggingface.co/datasets/emozilla/fineweb-10bt-tokenized-datatrove-llama2/resolve/main/00000_{}_shuffled.ds"         --start 0         --end 1         --left-pad-zeros 5
```

shared/modeling/examples/train.rs
```diff
@@ -233,38 +284,95 @@ async fn main() -> Result<()> {
...
+                Ok(dataset) => {
+                    info!(
+                        "Loaded local dataset with {} samples",
+                        dataset.num_sequences()
+                    );
+                    DataProvider::Local(dataset)
+                }
+                Err(err) => {
+                    println!(
+                        "Failed to load with local data provider. {err:?} Trying preprocessed data provider instead"
```
philrhc: in main, the train.rs falls back to pre-processed data if local data fails, this change keeps this behaviour, but might be more clear if it has an explicit option.

IAvecilla: Yeah, I think it’s a good idea to separate the possibility of using `Preprocessed` data. We might even end up splitting it further into Preprocessed Local (this behavior) and HTTP Preprocessed, which is the one I mentioned in the issue and whose implementation is here: [https://github.com/PsycheFoundation/psyche/pull/506](https://github.com/PsycheFoundation/psyche/pull/506). That should cover all the different data provider possibilities we have at the moment.

philrhc: please take a look at https://github.com/PsycheFoundation/nousnet/pull/511/commits/e479fd8ae66cc04a824fc6829791b3b36c2769bb
- adds local-preprocessed
- no longer falls back to preprocessed local data if the local data provider fails

PR #514, opened 2026-01-22
run-manager automatic run selection

dsocolobsky:
- If `RUN_ID` is not provided the run-manager will scan the solana RPC for those contracts owned by the Coordinator Program, list the runs and attempt to join one at random (which is not halted) (prioritizing waiting for compute runs).
- If `RUN_ID` is provided then the behavior is the same as usual and it attempts to join the specified run.
- Modified the user-facing docs to reflect these changes.
- Avoids joining runs which the user is not authorized to 
- Optional `--authorizer` param to only show runs where the given pubkey is the authorizer
- `list-runs` subcommand that lists all the runs and shows runs info

image

### DIFF
--- a/architectures/centralized/testing/src/lib.rs
+++ b/architectures/centralized/testing/src/lib.rs
@@ -3,7 +3,7 @@ pub mod server;
 pub mod test_utils;
 
 // Model Parameters
-pub const WARMUP_TIME: u64 = 60;
+pub const WARMUP_TIME: u64 = 20;
 pub const MAX_ROUND_TRAIN_TIME: u64 = 5;
 pub const ROUND_WITNESS_TIME: u64 = 2;
 pub const COOLDOWN_TIME: u64 = 3;
--- a/architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/instance_state.rs
+++ b/architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/instance_state.rs
@@ -191,6 +191,9 @@ impl CoordinatorInstanceState {
                 if !self.coordinator.model.check() {
                     return err!(ProgramError::ModelSanityCheckFailed);
                 }
+                if !self.coordinator.check_cold_start_warmup_steps() {
+                    return err!(ProgramError::ConfigSanityCheckFailed);
+                }
 
                 if self.coordinator.run_state == RunState::Uninitialized {
                     // this is the only way to get out of RunState::Uninitialized
@@ -310,6 +313,12 @@ impl CoordinatorInstanceState {
             }
 
             let _ = std::mem::replace(&mut self.coordinator.model, model);
+        }
+
+        if (config.is_some() || model.is_some())
+            && !self.coordinator.check_cold_start_warmup_steps()
+        {
+            return err!(ProgramError::ConfigSanityCheckFailed);
         }
 
         if let Some(progress) = progress {
--- a/architectures/decentralized/solana-tooling/tests/suites/memnet_coordinator_full_round.rs
+++ b/architectures/decentralized/solana-tooling/tests/suites/memnet_coordinator_full_round.rs
@@ -47,8 +47,8 @@ pub async fn run() {
     let join_authority = Keypair::new();
     let client = Keypair::new();
     let ticker = Keypair::new();
-    let warmup_time = 77;
-    let round_witness_time = 88;
+    let warmup_time = 10;
+    let round_witness_time = 10;
 
     // create the empty pre-allocated coordinator_account
     let coordinator_account = endpoint
@@ -97,7 +97,7 @@ pub async fn run() {
         Some(CoordinatorConfig {
             warmup_time,
             cooldown_time: 999,
-            max_round_train_time: 888,
+            max_round_train_time: 15,
             round_witness_time,
             min_clients: 1,
             init_min_clients: 1,
--- a/architectures/decentralized/solana-tooling/tests/suites/memnet_coordinator_rewards.rs
+++ b/architectures/decentralized/solana-tooling/tests/suites/memnet_coordinator_rewards.rs
@@ -93,7 +93,7 @@ pub async fn run() {
         Some(CoordinatorConfig {
             warmup_time,
             cooldown_time,
-            max_round_train_time: 888,
+            max_round_train_time: 15,
             round_witness_time,
             min_clients: 1,
             init_min_clients: 1,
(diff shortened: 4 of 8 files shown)


tools/rust-tools/run-manager/src/docker/coordinator_client.rs
```diff
@@ -40,6 +50,28 @@ impl CoordinatorClient {
...
+    fn fetch_run_state(&self, coordinator_account: &Pubkey) -> Option<RunState> {
+        match self.rpc_client.get_account(coordinator_account) {
+            Ok(coord_account) => match coordinator_account_from_bytes(&coord_account.data) {
+                Ok(account) => Some(account.state.coordinator.run_state),
+                Err(e) => {
+                    warn!(
+                        "Failed to deserialize coordinator account {}: {}",
+                        coordinator_account, e
+                    );
+                    None
```
dsocolobsky: This is too verbose for the user imho, will probably delete or leave as `debug!()`

tools/rust-tools/run-manager/src/docker/coordinator_client.rs
```diff
@@ -40,6 +50,28 @@ impl CoordinatorClient {
...
+            },
+            Err(e) => {
+                warn!(
+                    "Failed to fetch coordinator account {}: {}",
+                    coordinator_account, e
+                );
+                None
+            }
+        }
+    }
```
entropidelic: instead of having to do all this because you are returning an `Option`, just use `Result`:
```suggestion
        let coord_account = self.rpc_client.get_account(coordinator_account)?;
        let coord_account = coordinator_account_from_bytes(&coord_account.data)?;
        Ok(coord_account.state.coordinator.run_state)
```

the name of the variables could be better though, its a little confusing that everything is `account` or `coord_account`

tools/rust-tools/run-manager/src/docker/coordinator_client.rs
```diff
@@ -82,4 +114,57 @@ impl CoordinatorClient {
...
+                    self.program_id,
+                    e
+                )
+            })?;
+
+        let mut runs = Vec::new();
+        for (pubkey, account) in accounts {
+            match CoordinatorInstance::try_deserialize(&mut account.data.as_slice()) {
+                Ok(instance) => {
+                    if let Some(run_state) = self.fetch_run_state(&instance.coordinator_account) {
```
entropidelic: with the other suggestion I did, you just have to change this here
```suggestion
                    if let Ok(run_state) = self.fetch_run_state(&instance.coordinator_account) {
```

tools/rust-tools/run-manager/src/docker/manager.rs
```diff
@@ -64,13 +66,39 @@ impl RunManager {
...
+                }
+
+                let selected = select_best_run(&runs)?;
+                info!(
+                    "Selected run: {} (state: {})",
+                    selected.run_id, selected.run_state
+                );
+                selected.run_id.clone()
+            }
+        };
```
entropidelic:
```suggestion
        if let Ok(run_id) = std::env::var("RUN_ID") {
            info!("Using RUN_ID from environment: {}", run_id);
            return Ok(Self {
                wallet_key,
                run_id,
                coordinator_client,
                env_file,
                local_docker,
                scratch_dir,
            });
        }
        info!("RUN_ID not set, discovering available runs...");
        let runs = coordinator_client.get_all_runs()?;
        if runs.is_empty() {
            bail!("No runs found on coordinator program");
        }

        // Log all discovered runs
        info!("Discovered {} run(s):", runs.len());
        for run in &runs {
            info!("  - {} (state: {})", run.run_id, run.run_state);
        }

        let selected = select_best_run(&runs)?;
        info!(
            "Selected run: {} (state: {})",
            selected.run_id, selected.run_state
        );

        Ok(Self {
            wallet_key,
            run_id: selected.run_id.clone(),
            coordinator_client,
            env_file,
            local_docker,
            scratch_dir,
        })
```

tools/rust-tools/run-manager/src/docker/coordinator_client.rs
entropidelic: or maybe you can do a `match` and log the error

tools/rust-tools/run-manager/src/docker/manager.rs
dsocolobsky: thx

tools/rust-tools/run-manager/src/docker/manager.rs
```diff
@@ -320,3 +356,93 @@ impl RunManager {
...
+
+/// Parse wallet key string to extract the user's pubkey.
+fn parse_wallet_pubkey(wallet_key: &str) -> Result<Pubkey> {
+    let keypair = if wallet_key.starts_with('[') {
+        // Assume Keypair::read format (JSON array of bytes)
+        Keypair::read(&mut Cursor::new(wallet_key))
+            .map_err(|e| anyhow!("Failed to parse wallet key: {}", e))?
+    } else {
+        // Assume base58 encoded private key
+        Keypair::from_base58_string(wallet_key)
```
entropidelic: isn't this fallible? what happens if the string is not really a base58 encoded private key?

tools/rust-tools/run-manager/src/docker/manager.rs
```diff
@@ -320,3 +356,93 @@ impl RunManager {
...
+
+    if candidates.is_empty() {
+        bail!(
+            "No joinable runs found. All {} run(s) are in unjoinable states.",
+            runs.len()
+        );
+    }
+
+    // Filter by join_authority if --authorizer was specified
+    if let Some(auth) = authorizer {
```
entropidelic: what is the behavior when `--authorizer` is not set?

dsocolobsky: It just shows all the runs you are allowed to join. `--authorizer` is a filter by runs with a certain authorizer (e.g. if we want to filter only by Nous runs).

tools/rust-tools/run-manager/src/docker/manager.rs
dsocolobsky: Yeah `from_base58_string` has an internal `unwrap()` apparently

But I found an alternative approach calling two functions and we can validate first with a nicer error message, just pushed the changes.

tools/rust-tools/run-manager/src/docker/manager.rs
```diff
@@ -320,3 +357,173 @@ impl RunManager {
...
+            Ok(grantee)
+        }
+        None => {
+            bail!(
+                "User {} is not authorized to join run {}",
+                user_pubkey,
+                run_id
+            );
+        }
+    }
```
entropidelic:
```suggestion
    let Some(grantee) = coordinator_client.can_user_join_run(run_id, user_pubkey, delegate_authorizer)? else {
        bail!(
            "User {} is not authorized to join run {}",
            user_pubkey,
            run_id
        );
   };
        
   info!("Resolved AUTHORIZER={} for run {}", grantee, run_id);
   Ok(grantee)
}
```

PR #516, opened 2026-01-23
Add `parallelism_auto` flag to automatically set dp, tp and micro batch size

pefontana:
Adds --parallelism-auto flag to automatically detect optimal dp, tp, and micro_batch_size based on the model and GPU hardware.

  1. Detects GPU type via nvidia-smi (with fallback to /proc/driver/nvidia/gpus/*/information for containers)
  2. Fetches parallelism_data.json from the model's HuggingFace repo
  3. Looks up config: GPU type → num GPUs → {dp, tp, micro_batch_size}

```
  {
    "H100": {
      "1": { "dp": 1, "tp": 1, "micro_batch_size": 4 },
      "8": { "dp": 4, "tp": 2, "micro_batch_size": 4 }
    },
    "H200": {
      "8": { "dp": 8, "tp": 1, "micro_batch_size": 8 }
    }
  }
```

We considered sharing the parallelism config via P2P but it added unnecessary complexity—dp/tp/micro_batch_size are needed before model download begins, requiring a separate request/wait flow. Since the file is ~200 bytes, fetching from HF is fast and keeps the code simple.

### Note:
I think we can merge this PR now, we should only take into account that if the run is private, is posible that some trainers dont have access to HuggingFace, so I dont recommend using this flag for private runs now
Once #455 is merged, all trainers must have access to HF/GCP, so it wont be a problem

### DIFF
--- a/.github/workflows/build-run-manager.yml
+++ b/.github/workflows/build-run-manager.yml
@@ -23,6 +23,7 @@ jobs:
     # If we use ubuntu-latest it will link against a newer version of glibc making
     # the binary non-portable to our 22.04 servers, so use this for compatibility.
     runs-on: ubuntu-22.04
+    timeout-minutes: 60
 
     steps:
       - name: Checkout repository
--- a/.github/workflows/push-docker.yml
+++ b/.github/workflows/push-docker.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   push-docker-image:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
       - name: Free Disk Space (Ubuntu)
         uses: jlumbroso/free-disk-space@main
--- a/.github/workflows/run-manager-integration-test-base.yml
+++ b/.github/workflows/run-manager-integration-test-base.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   run-manager-integration-test:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     permissions:
       contents: read
       packages: write
--- a/.github/workflows/solana-build-anchor-programs.yml
+++ b/.github/workflows/solana-build-anchor-programs.yml
@@ -6,6 +6,7 @@ on:
 jobs:
   build-validator-image:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
       - uses: actions/checkout@v4
         with:
--- a/.github/workflows/solana-integration-test-base.yml
+++ b/.github/workflows/solana-integration-test-base.yml
@@ -11,6 +11,7 @@ on:
 jobs:
   solana-integration-test:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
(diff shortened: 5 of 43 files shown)


shared/client/src/parallelism_lookup.rs
```diff
@@ -0,0 +1,85 @@
...
+    pub dp: usize,
+    pub tp: usize,
+    pub micro_batch_size: usize,
+}
+
+type Table = HashMap<String, HashMap<String, HashMap<String, ParallelismConfig>>>;
+
+fn get_gpu_type() -> String {
+    // Try nvidia-smi first
+    let raw = Command::new("nvidia-smi")
```
entropidelic: small nit: maybe we can rename this variable to something better?

shared/client/src/parallelism_lookup.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    // Normalize GPU name to match table keys
+    if raw_gpu_name.to_uppercase().contains("H200") {
+        "H200".to_string()
+    } else if raw_gpu_name.to_uppercase().contains("H100") {
+        "H100".to_string()
+    } else {
+        raw_gpu_name
+    }
+}
```
ethernet8023: have you considered using the `nvml_wrapper` crate instead of shelling out / reading /proc/fs stuff? we can grab gpu count from there too 🤷 and assert that they're all the same GPU for sanity checking :D

we use this in some of the metrics stuff already - 
it would be something like:
```rs
use nvml_wrapper::Nvml;

#[derive(Debug)]
struct GpuInfo {
    name: String,
    device_count: u32,
}

fn get_gpu_info() -> anyhow::Result<GpuInfo> {
    let nvml = Nvml::init()?;
    let device_count = nvml.device_count()?;

    if device_count == 0 {
        anyhow::bail!("No GPUs found!");
    }

    let mut gpu_names = Vec::new();

    for i in 0..device_count {
        let device = nvml.device_by_index(i)?;
        gpu_names.push(device.name()?);
    }

    let first_name = &gpu_names[0];
    if !gpu_names.iter().all(|name| name == first_name) {
        anyhow::bail!(
            "All GPUs must be of the same type, but we have mismatching names: {:?}",
            gpu_names
        );
    }

    Ok(GpuInfo {
        name: gpu_names.pop().unwrap(),
        device_count,
    })
}
```

pefontana: Thanks Ari!
I add it and works

shared/client/src/parallelism_lookup.rs
pefontana: done!

shared/client/src/parallelism_lookup.rs
pefontana: I just changed the `let device_count = nvml.device_count()?;` because was not taking into account `CUDA_VISIBLE_DEVICES` env var

PR #519, opened 2026-01-25
VLLM nix packaging second attempt!

ethernet8023:
passes both train test and compiling cudnn stuff on h100

- add vllm package and update the world

- fix some cuda lib path problems in vllm
- updates mdbook
- fix some clippy lints on new versions

- refactor nix building stuff to build examples in CI using nix, moving closer to unification of build processes
- more nix work to approach splitting vllm out of packages that don't need it, for later

later, we should consider splitting the vllm code into a separate set of deps, so the regular psyche binaries don't get vllm included

Original PR was https://github.com/PsycheFoundation/psyche/pull/464

### DIFF
--- a/nix/nixpkgs.nix
+++ b/nix/nixpkgs.nix
@@ -32,10 +32,6 @@ in
             liger-kernel = pyfinal.callPackage ../python/liger-kernel.nix { };
             torchtitan = pyfinal.callPackage ../python/torchtitan.nix { };
             torchdata = pyfinal.callPackage ../python/torchdata.nix { };
-            keras = pyprev.keras.overrideAttrs (oldAttrs: {
-              dontUsePytestCheck = "please dont";
-
-            });
           };
         };
       })


nix/python.nix
```diff
@@ -31,6 +31,7 @@ let
   # packages that we provide to the venv via nix derivations
   topLevelNixPkgs = [
     "torch"
+    "vllm"
```
IAvecilla: I don't know if we want to support vllm for `aarch64-darwin` but I think this is not compiling on that architecture:
```
error: Package ‘python3.12-vllm-0.14.0’ in /nix/store/m6ji2a8jp7w4ck7vvrjh4grs4p3n7r26-source/pkgs/development/python-modules/vllm/default.nix:553 is not available on the requested hostPlatform:
         hostPlatform.system = "aarch64-darwin"
```
Maybe we can move inside the `isLinux` check

ethernet8023: done - also made the inference node packages not build at all on macos since they won't work without vllm

arilotter:
oh my god the tests passed im gonna cry :)

arilotter:
verified this works after merging with https://github.com/PsycheFoundation/psyche/pull/487

PR #525, opened 2026-01-27
Move Coordinator metadata off-chain

IAvecilla:


shared/client/src/state/init.rs
```diff
@@ -199,11 +206,56 @@ impl<T: NodeIdentity, A: AuthenticatableIdentity + 'static> RunInitConfigAndIO<T
...
+                    )
+                    .await
+                    {
+                        Ok(config) => config,
+                        Err(e) => {
+                            error!(
+                                "Failed to fetch model extra data from Hub ({}), using default: {}",
+                                repo_id, e
+                            );
+                            return Err(InitRunError::GcsModelLoad(e));
```
entropidelic: i think this is returning `GcsModelLoad` error and should be `HubModelLoad`. Also, is the error log correct? not sure if the default log is ok

PR #535, opened 2026-02-02
Improve authorization checks and retries

IAvecilla:
Mainly two improvements are introduced with this PR:
- The message for the error `AccountNotInitialized` was confusing and we always keep retrying to join a run where the client does not have authorization. This will prevent the automatic retry and also make the error explicit
- Now the `can_join` command and the `train` will both try to join the run as it was permissionless (using the system program as default) and fallback to the `--authorizer` pubkey or if not provided try to use the pubkey of the user running the command. This is just to avoid always using the systems program in case someone forgots to use the `authorizer` flag


IAvecilla:
All this new logic is being added to #514 since it checks authorization previous to join the run and is already improving the error messages

Issue #539, opened 2026-02-03
fix validate-config


arilotter:
it passes validation for some state.toml files that don't actually work when loaded in D:

PR #544, opened 2026-02-03
fix: validate-config was not throwing an exit code

ethernet8023:
closes https://github.com/PsycheFoundation/psyche/issues/539

### DIFF
--- a/config/consilience/state.toml
+++ b/config/consilience/state.toml
@@ -1,7 +1,10 @@
+run_id = "consilience-40b"
+run_state = "WaitingForMembers"
+
 [config]
 warmup_time = 3600
 cooldown_time = 120
-rounds_per_epoch = 500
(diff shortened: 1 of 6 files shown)


nix/checks.nix
```diff
@@ -59,16 +59,22 @@
                   exit 1
                 fi
 
+
                 for f in $dir/*; do
                   if [ -f $f/data.toml ]; then
+                  echo "ccecking $(realpath -s --relative-to $dir $f/data.toml) and $(realpath -s --relative-to $dir $f/state.toml)"
```
pefontana, line 65: just a typo

```suggestion
                  echo "checking $(realpath -s --relative-to $dir $f/data.toml) and $(realpath -s --relative-to $dir $f/state.toml)"
```

pefontana (approved):

PR #546, opened 2026-02-04
Website backend basic deployment pipeline to Railway

rewbs:
This adds a github workflow to build the website backend docker image, push it to dockerhub, and deploy it as a Railway app.

The deployment topology is a "lift and shift" of current code (combined API and indexer process, filesystem-backed) and is not intended to be as-is in the long run, but can be run in an HA deployment (multiple load balanced instances) with the following caveats:
- Max ~2.5k concurrent website viewers per railway instance.
- Each node generates full load on solana RPC (rate limit would need to be adjusted based on railway instance count).
- Users will see slightly different data depending on which railway instance handles their request (because each instance may have a different view of the data).

Example successful run of the workflow: https://github.com/PsycheFoundation/nousnet/actions/runs/21651181317

For more info on ideas to improve the deployment topology in the future, see: https://discord.com/channels/1173234175801827380/1308967411025973278/1466767012993831096

### DIFF
--- a/nix/docker.nix
+++ b/nix/docker.nix
@@ -224,6 +224,7 @@ let
         Env = [
           "NODE_ENV=production"
           "STATE_DIRECTORY=/data"
+          "SSL_CERT_FILE=/etc/ssl/certs/ca-bundle.crt"
         ];
         Cmd = [ "/bin/backend" ];
         ExposedPorts = {


rewbs:
Ignore this - wrong indexer. :D

PR #548, opened 2026-02-04
upgrade to iroh 0.96

ethernet8023:
network: add iroh metrics

### DIFF
--- a/shared/network/src/router.rs
+++ b/shared/network/src/router.rs
@@ -137,34 +137,14 @@ mod tests {
                         .address_lookup(static_discovery.clone())
                         .bind()
                         .await?;
-                    let blobs = MemStore::new();
                     let gossip = Gossip::builder().spawn(endpoint.clone());
-                    let (tx_model_parameter_req, _rx_model_parameter_req) =
-                        tokio::sync::mpsc::unbounded_channel();
-                    let (tx_model_config_req, _rx_model_parameter_req) =
-                        tokio::sync::mpsc::unbounded_channel();
-                    let p2p_model_sharing =
-                        ModelSharing::new(tx_model_parameter_req, tx_model_config_req);
-                    let blobs_protocol = BlobsProtocol::new(&blobs.clone(), None);
 
-                    let allowlist_clone = allowlist.clone();
-                    let allowlisted_blobs = AccessLimit::new(blobs_protocol, move |endpoint_id| {
-                        allowlist_clone.allowed(endpoint_id)
-                    });
-                    let allowlist_clone_2 = allowlist.clone();
                     let allowlisted_gossip = AccessLimit::new(gossip.clone(), move |endpoint_id| {
-                        allowlist_clone_2.allowed(endpoint_id)
+                        allowlist.allowed(endpoint_id)
                     });
-                    let allowlist_clone_3 = allowlist.clone();
-                    let allowlisted_model_sharing =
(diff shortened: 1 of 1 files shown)


shared/network/src/connection_monitor.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub connection_type: ConnectionType,
+    pub latency: Duration,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ConnectionType {
+    Direct,
+    Relay,
+    Mixed,
+}
```
IAvecilla: I think we already have some implementations of the `ConnectionType` in `network/lib.rs` and `metrics/lib.rs`. Not really a priority right now if it gets complicated to reuse those but maybe mark some of them as a todo for future removal

shared/network/src/state.rs
```diff
@@ -66,6 +66,7 @@ impl BandwidthTracker {
         }
     }
 
+    #[allow(dead_code)]
```
IAvecilla: I think we're using this in `shared/network/src/lib.rs`

shared/network/src/lib.rs
```diff
@@ -320,6 +347,35 @@ where
...
+            let builder = iroh_n0des::Client::builder(&endpoint);
+            let allowlist = allowlist.clone();
+            (async move {
+                let client = builder.api_secret_from_env()?.build().await?;
+                const API_SECRET_ENV_VAR_NAME: &str = "N0DES_API_SECRET";
+
+                match std::env::var(API_SECRET_ENV_VAR_NAME) {
+                    Ok(ticket_string) => {
+                        let ticket = ApiSecret::from_str(&ticket_string)
+                            .context("invalid {API_SECRET_ENV_VAR_NAME}")?;
```
IAvecilla: I think we need to do ` .context(format!("invalid {API_SECRET_ENV_VAR_NAME}"))` for this to work as expected

shared/network/Cargo.toml
```diff
@@ -34,6 +34,7 @@ tokenizers.workspace = true
 get_if_addrs = "0.5.3"
 n0-future = "0.2.0"
```
IAvecilla: Do we need to bump this version too? I think iroh uses `0.3.1` but not sure if it will be a problem

shared/network/src/connection_monitor.rs
```diff
@@ -0,0 +1,172 @@
...
+                    );
+
+                    {
+                        let mut conns = connections.write().unwrap();
+                        conns.insert(remote_id, ConnectionData {
+                            endpoint_id: remote_id,
+                            connection_type: conn_type,
+                            latency,
+                        });
+                    }
```
IAvecilla, line 82: Just as a sanity check: I think we only update this info after the handshake, but we’re not updating it over time. Is that expected? In `extract_connection_info` we have the `paths_watcher`, which could give us updates over time, but we’re dropping it after the connection is established for the first time. If we only want a quick estimate, that’s fine, but I’m not sure whether we want to track this over time

shared/network/src/connection_monitor.rs
ethernet8023: removed in favor of the metrics one

shared/network/src/connection_monitor.rs
ethernet8023: oops, good point. i check it periodically now :)

PR #549, opened 2026-02-05
feat: special claimer wallet

crypto-vincent:
## Summary

Add a "claimer" pubkey to the clients that can be used to claim the rewards from the treasurer instead of using the client's signer to sign during treasurer claims.

## Details

TLDR:

- the `Coordinator.join_run` IX now takes a `claimer` as parameter
- the `Treasurer.participant_claim` now requires the `claimer` to sign instead of the user (the client)
- the on-chain storage now needs a breaking change
- the CLI must then be upgraded
- smart contract tests updated to reflect new logic


architectures/decentralized/solana-coordinator/programs/solana-coordinator/src/client.rs
```diff
@@ -25,6 +25,8 @@ use ts_rs::TS;
 #[ts(rename = "SolanaClient")]
 pub struct Client {
     pub id: ClientId,
+    #[ts(type = "number[]")]
+    pub claimer: Pubkey,
```
crypto-vincent, line 30: this is the main change

PR #555, opened 2026-02-09
Add initial upload/download commands to run manager

benbarclay:
### The Change

Add `upload_data` and `download_results` commands to run-manager CLI.

`upload_data` is intended for uploading initial run data. This is, to my knowledge, currently not specced out, but required. This command is intended to be the basis of that implementation, but it's likely that either the command or the backing service will require changes as the requirement is fully specced out. It supports single file or directory uploads.

`download_results` is intended for downloading the final results of a run, which have been uploaded to a Google Cloud Storage Bucket.

Currently these are both pointing at the same bucket (controlled by the `run-down` service), which is likely something we will want to change in production. There's likely a few things like that which will need to be changed, but they can mostly be handled by `run-down`. Still, probably best to view this as an initial implementation rather than the expected final result, as there's still a few integration pieces missing.

Both of these services authenticate to `run-down` via solana wallet private key auth. The commands generate a signed message and attach the signature to the request as a header. `run-down` will pull the correct public key from the indexer and validate the signature with that. One quirky note is that we never actually transmit the signed message, but all the contents are included in the request - allowing the server to reconstruct the message and validate there.

#### Upload a single file

run-manager upload-data --run-id <RUN_ID> --path /path/to/file.txt --wallet-private-key-path ~/.config/solana/id.json

#### Upload a directory (preserves structure)

run-manager upload-data --run-id <RUN_ID> --path /path/to/directory --wallet-private-key-path ~/.config/solana/id.json

#### Download results

run-manager download-results --run-id <RUN_ID> --output-dir ./results  --wallet-private-key-path ~/.config/solana/id.json

### DIFF
--- a/tools/rust-tools/run-manager/src/commands/run/download_results.rs
+++ b/tools/rust-tools/run-manager/src/commands/run/download_results.rs
@@ -1,25 +1,15 @@
 use std::path::PathBuf;
 
+use crate::commands::Command;
 use anyhow::{Context, Result, bail};
 use async_trait::async_trait;
 use clap::Args;
-use serde::{Deserialize, Serialize};
-
-use crate::commands::Command;
+use futures::TryStreamExt;
 use psyche_solana_rpc::SolanaBackend;
+use tokio::fs::File;
+use tokio_util::io::StreamReader;
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
-struct DownloadUrlEntry {
-    path: String,
-    url: String,
-}
-
-#[derive(Debug, Clone, Serialize, Deserialize)]
-struct DownloadUrlsResponse {
-    urls: Vec<DownloadUrlEntry>,
-    #[serde(rename = "expiresAt")]
-    expires_at: String,
-}
+use super::run_down_service;
 
 #[derive(Debug, Clone, Args)]
 #[command()]
@@ -32,6 +22,9 @@ pub struct CommandDownloadResults {
 
     #[clap(long, env, default_value = "3600")]
     pub expires_in_seconds: u64,
+
+    #[clap(long)]
+    pub overwrite: bool,
 }
 
 #[async_trait]
@@ -41,63 +34,58 @@ impl Command for CommandDownloadResults {
             run_id,
             output_dir,
             expires_in_seconds,
+            overwrite,
         } = self;
 
-        // Generate a random nonce
-        let nonce: u64 = rand::random();
+        // Check if output directory exists and is not empty
+        if output_dir.exists() && !overwrite {
+            let mut entries = tokio::fs::read_dir(&output_dir)
+                .await
+                .context("Failed to read output directory")?;
 
-        // Create the message to sign: ${runId}:${expiresInSeconds}:${nonce}
-        let message = format!("{}:{}:{}", run_id, expires_in_seconds, nonce);
-        let message_bytes = message.as_bytes();
+            if entries.next_entry().await?.is_some() {
+                println!(
+                    "Warning: Output directory {:?} already exists and contains files.",
+                    output_dir
+                );
+                println!("Files may be overwritten during download.");
+                print!("Continue? [y/N]: ");
+                std::io::Write::flush(&mut std::io::stdout())?;
 
-        // Sign the message using the backend's wallet
-        let signature = backend.sign_message(message_bytes);
+                let mut response = String::new();
+                std::io::stdin().read_line(&mut response)?;
 
-        // Encode the signature in base58
-        let signature_b58 = bs58::encode(&signature).into_string();
-
-        // Create the request body
-        #[derive(Serialize)]
-        struct RequestBody {
-            #[serde(rename = "expiresInSeconds")]
-            expires_in_seconds: u64,
-            nonce: String,
+                if !response.trim().eq_ignore_ascii_case("y") {
+                    println!("Download cancelled.");
+                    return Ok(());
+                }
+            }
         }
-
-        let request_body = RequestBody {
-            expires_in_seconds,
-            nonce: nonce.to_string(),
-        };
-
-        // Make POST request to the API
-        let api_url = format!("https://run-down.nousresearch.com/v1/download/{}", run_id);
-
-        let client = reqwest::Client::new();
-        let response = client
-            .post(&api_url)
-            .header("X-Solana-Signature", signature_b58)
-            .json(&request_body)
-            .send()
-            .await
-            .context("Failed to fetch download URLs")?;
-
-        if !response.status().is_success() {
-            let status = response.status();
-            let error_text = response.text().await.unwrap_or_default();
-            bail!("API request failed with status {}: {}", status, error_text);
-        }
-
-        let urls_response: DownloadUrlsResponse = response
-            .json()
-            .await
-            .context("Failed to parse response JSON")?;
-
-        println!("Found {} files to download", urls_response.urls.len());
 
         // Create output directory if it doesn't exist
         tokio::fs::create_dir_all(&output_dir)
             .await
             .context("Failed to create output directory")?;
+
+        // Generate a random nonce
+        let nonce: u64 = rand::random();
+
+        // Generate signature for the run-down service
+        let signature_b58 =
+            run_down_service::generate_signature(&backend, &run_id, expires_in_seconds, nonce);
+
+        // Make POST request to the API
+        let client = reqwest::Client::new();
+        let urls_response = run_down_service::get_download_urls(
+            &client,
+            &run_id,
+            &signature_b58,
+            expires_in_seconds,
+            nonce,
+        )
+        .await?;
+
+        println!("Found {} files to download", urls_response.urls.len());
 
         // Download each file
         for (idx, entry) in urls_response.urls.iter().enumerate() {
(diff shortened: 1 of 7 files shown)


tools/rust-tools/run-manager/src/commands/run/download_results.rs
```diff
@@ -0,0 +1,155 @@
...
+
+            // Preserve the directory structure from the path
+            let file_path = output_dir.join(&entry.path);
+
+            // Create parent directories if they don't exist
+            if let Some(parent) = file_path.parent() {
+                tokio::fs::create_dir_all(parent)
+                    .await
+                    .with_context(|| format!("Failed to create directory {:?}", parent))?;
+            }
```
dsocolobsky, line 121: I'm not 100% sure but here I believe if we do not sanitize we should be careful with absolute or relative paths since they could write outside the project folder.

i.e. if path is `../bashrc` or something it could overwrite ~/.bashrc

afaik we could do something like this

```rust
  let sanitized = Path::new(&entry.path)
      .components()
      .filter(|c| matches!(c, std::path::Component::Normal(_)))
      .collect::<PathBuf>();
  let file_path = output_dir.join(sanitized);
```

tools/rust-tools/run-manager/src/commands/run/upload_data.rs
```diff
@@ -0,0 +1,194 @@
...
+
+            let upload_response: UploadUrlResponse = response
+                .json()
+                .await
+                .context("Failed to parse upload URL response")?;
+
+            // Read the file contents
+            let file_contents = fs::read(file_path)
+                .await
+                .with_context(|| format!("Failed to read file: {:?}", file_path))?;
```
entropidelic: maybe we can check the contents before doing the initial request? So if there is some error we catch it before doing the request

tools/rust-tools/run-manager/src/commands/run/download_results.rs
```diff
@@ -0,0 +1,155 @@
...
+        // Sign the message using the backend's wallet
+        let signature = backend.sign_message(message_bytes);
+
+        // Encode the signature in base58
+        let signature_b58 = bs58::encode(&signature).into_string();
+
+        // Create the request body
+        #[derive(Serialize)]
+        struct RequestBody {
+            #[serde(rename = "expiresInSeconds")]
```
ethernet8023: you can throw `#[serde(rename_all = "camelCase")]` on the struct itself to get js-style camelcase attrs :)

tools/rust-tools/run-manager/src/commands/run/download_results.rs
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        let urls_response: DownloadUrlsResponse = response
+            .json()
+            .await
+            .context("Failed to parse response JSON")?;
+
+        println!("Found {} files to download", urls_response.urls.len());
+
+        // Create output directory if it doesn't exist
```
ethernet8023: maybe try this first, so a permission error on a dir won't make a request / signature?

tools/rust-tools/run-manager/src/commands/run/upload_data.rs
```diff
@@ -0,0 +1,194 @@
...
+        if path.is_file() {
+            files.push(path);
+        } else if path.is_dir() {
+            let mut sub_files = Box::pin(collect_files_from_dir(&path)).await?;
+            files.append(&mut sub_files);
+        }
+    }
+
+    Ok(files)
+}
```
entropidelic, line 152: there is a rust crate called WalkDir that is widely used and should be the perfect fit for this. In fact we already have it as a transitive dependency in the project so it should not bring anything new 👍 
https://github.com/BurntSushi/walkdir

tools/rust-tools/run-manager/src/commands/run/download_results.rs
```diff
@@ -0,0 +1,155 @@
...
+            let file_path = output_dir.join(&entry.path);
+
+            // Create parent directories if they don't exist
+            if let Some(parent) = file_path.parent() {
+                tokio::fs::create_dir_all(parent)
+                    .await
+                    .with_context(|| format!("Failed to create directory {:?}", parent))?;
+            }
+
+            let bytes = file_response
```
ethernet8023: given that we're using async & tokio here, rather than downloading the entire file into memory & then writing the whole thing to disk before hitting the next file, we can overlap disk I/O and network I/O with something like

```rs
use futures::TryStreamExt;
use tokio_util::io::StreamReader;

let mut file = File::create(&file_path).await.with_context(|| format!("Failed to create file {file_path:?}"))?;
let stream = file_response.bytes_stream();
tokio::io::copy(&mut StreamReader::new(stream), &mut file)
	.await
	.with_context(|| format!("Failed to download to file {file_path:?}"))?
```
after enabling the `stream` feature on `reqwest`

tools/rust-tools/run-manager/src/commands/run/download_results.rs
```diff
@@ -0,0 +1,155 @@
...
+            run_id,
+            output_dir,
+            expires_in_seconds,
+        } = self;
+
+        // Generate a random nonce
+        let nonce: u64 = rand::random();
+
+        // Create the message to sign: ${runId}:${expiresInSeconds}:${nonce}
+        let message = format!("{}:{}:{}", run_id, expires_in_seconds, nonce);
```
ethernet8023: a small design consideration; in theory, if some other solana program uses something similar to this format, you could use this signature for Evil - what if we added a prefix to the message, something like `nous-run-down-service:` to make the possibility of it matching up much smaller?

tools/rust-tools/run-manager/src/commands/run/download_results.rs
ethernet8023: what about a simple overwrite check if the path / any of the paths exist?
```
this will overwrite existing file /absolute/path/to/file.txt
continue? [y/N] 
```

tools/rust-tools/run-manager/src/commands/run/upload_data.rs
```diff
@@ -0,0 +1,194 @@
...
+                idx + 1,
+                files_to_upload.len(),
+                relative_path
+            );
+
+            // Generate a random nonce for this file
+            let nonce: u64 = rand::random();
+
+            // Create the message to sign: ${runId}:${expiresInSeconds}:${nonce}
+            let message = format!("{}:{}:{}", run_id, expires_in_seconds, nonce);
```
ethernet8023: maybe extract the message/signature creation into a fn so we can share this between upload/download? seems identical between up/down

tools/rust-tools/run-manager/src/commands/run/upload_data.rs
```diff
@@ -0,0 +1,194 @@
...
+            // Read the file contents
+            let file_contents = fs::read(file_path)
+                .await
+                .with_context(|| format!("Failed to read file: {:?}", file_path))?;
+
+            let file_size = file_contents.len();
+            println!("  Uploading {} bytes...", file_size);
+
+            // Upload the file to the signed URL
+            let upload_response = client
```
ethernet8023, line 116: you can stream the body instead of reading first then sending using 
```rs
let file = File::open(file_path).await.with_context(|| format!("Failed to read file: {:?}", file_path))?;

let upload_response = client
	.put(&upload_response.url)
	.header("Content-Type", "application/octet-stream")
	.body(file)
	.send()
	.await
	.context("Failed to upload file to signed URL")?;
```

tools/rust-tools/run-manager/src/commands/run/upload_data.rs
benbarclay: Yeah, I was thinking about that but wasn't really sure where to stick the extracted logic. Eh, I'll find a place for it.

tools/rust-tools/run-manager/src/commands/run/download_results.rs
benbarclay: That's a good call, though that only stops accidental re-use, rather than deliberate malicious use. I'll add it in.

tools/rust-tools/run-manager/src/commands/run/download_results.rs
benbarclay: Neat!

tools/rust-tools/run-manager/src/commands/run/download_results.rs
benbarclay: Can add a `--no-preserve-root` flag :P

It's a good call though. I don't think that the concept of a project folder is as useful here due to eventual tool being a CLI that can run anywhere, but I do like the idea of adding in an overwrite check.

tools/rust-tools/run-manager/src/commands/run/upload_data.rs
benbarclay: Hmm, I like the idea, but it also seems vaguely opposed to Ari's suggestion to do streaming below, rather than opening the file 🤔

tools/rust-tools/run-manager/src/commands/run/run_down_service.rs
benbarclay, line 1: So this contains some extracted generic functionality - but I don't really have a great place to stick this file. Thoughts @arilotter ?

tools/rust-tools/run-manager/src/commands/run/download_results.rs
benbarclay: This is updated both here and in the run-down service

tools/rust-tools/run-manager/src/commands/run/run_down_service.rs
ethernet8023: i think that as long as we're only using it inside run-manager, this makes sense to leave here. later, if we use this functionality in another crate, we could drop a new crate for run-down functionality in `shared/`

PR #556, opened 2026-02-09
feat: add --permissionless flag for authorization commands

### LINKED ISSUE #471: Permissionless authorization flag
To make a permissionless run, we currently have to issue an authorization with the authorize set to `11111111111111111111111111111111`, which is a bit obscure for people not very familiar with Solana. It would be better to create a new `--permissionless` flag that uses this address under the hood and conflicts with the existing `--authorize` flag, so users can use one or the other.. Here's a comment with the [https://github.com/PsycheFoundation/psyche/pull/457#discussion_r2665901308](https://github.com/PsycheFoundation/psyche/pull/457#discussion_r2665901308)

PivasDesant:
Adds `--permissionless` flag to authorization commands as an alternative to `--authorizer 11111111111111111111111111111111`.

The flag automatically uses `system_program::ID` internally and conflicts with `--authorizer` to prevent misuse.

Affected commands:
- `join-authorization-create`
- `join-authorization-delete`
- `join-authorization-read`
- `can-join`

Closes #471

### DIFF
--- a/tools/rust-tools/run-manager/src/commands/authorization/create.rs
+++ b/tools/rust-tools/run-manager/src/commands/authorization/create.rs
@@ -4,8 +4,8 @@ use anyhow::Result;
 use async_trait::async_trait;
 use clap::Args;
 
-use psyche_solana_rpc::instructions;
 use psyche_solana_rpc::SolanaBackend;
+use psyche_solana_rpc::instructions;
 
 #[derive(Debug, Clone, Args)]
 #[command()]


arilotter:
Thank you for the PR!!! this is sweet, almost exactly what i need. now that i see it down in a PR,  I wonder if we shouldn't generally replace the authorization stuff in the CLIs with an enum, like
```rs
enum Authorization {
  Address(..),
  Permissionless
 }
 ```
 
 and then can implement this in such a way that if you pass `--authorization permissionless` that maps to the 11111111 addr.... @crypto-vincent what do you think?

PR #560, opened 2026-02-10
try fix CI tests

dsocolobsky:
Tests were not passing in the CI, mostly due to timing issues.

- We waited 600 seconds for retry in subscriptions, lowered it to 30 seconds
- 30s->60s wait in p2p test, this seems to fix it, apparently couldn't find the peer because of timing issues in the CI
- In disconnect_client we kill the client in step 2 instead of step 10. Apparently we weren't reaching step 10 and it got stuck in a loop.
- relax some checks in the same test, check for at least 1 healthcheck not neccesarily 2.
- show logs of containers on crash

### DIFF
--- a/shared/network/src/router.rs
+++ b/shared/network/src/router.rs
@@ -134,6 +134,7 @@ mod tests {
                     let static_discovery = MemoryLookup::new();
                     let endpoint = Endpoint::builder()
                         .secret_key(k)
+                        .clear_address_lookup()
                         .address_lookup(static_discovery.clone())
                         .bind()
                         .await?;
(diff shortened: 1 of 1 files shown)


architectures/decentralized/solana-common/src/backend.rs
```diff
@@ -123,6 +123,14 @@ async fn subscribe_to_account(
                         }
                     }
                 }
+                _ = tokio::time::sleep(Duration::from_secs(30)) => {
```
pefontana: mmm I think I dont like this.
Tell me if i am correct, but wont this kill the Solana subscription if we dont receive any updates in 30 seconds?
That can easily happen in warmup or cooldown state

architectures/decentralized/testing/tests/integration_tests.rs
```diff
@@ -410,13 +410,10 @@ async fn disconnect_client() {
                 }
 
                 if epoch == 0
-                    && step == 10
+                    && step == 2
                     && old_state == RunState::RoundTrain.to_string()
                     && !killed_client
                 {
-                    let epoch_clients = solana_client.get_current_epoch_clients().await;
```
pefontana, line 417: Just a question, not sure if it is really important, but Is necessary to remove this check?

architectures/decentralized/solana-common/src/backend.rs
dsocolobsky: yeah we can remove it, I added to test some changes but apparently it's not needed

architectures/decentralized/testing/tests/integration_tests.rs
dsocolobsky: yeah we can remove it, I added to test some changes but apparently it's not needed

PR #563, opened 2026-02-10
fix inconsistent tests which are flaky in the CI

dsocolobsky:
Can't guarantee but I think the tests in the CI now run somewhat better.
All the changes are test related, none in the core logic:

- witness time 1s -> 5s to give time for clients to send their witness message
- cooldown time 30s -> 15s
- in the pause_and_resume test we now wait until epoch 2 not until epoch 5 to pause, to make it run faster
- adjusted some other times and waits to play nicely with the witness time increase
- adjusted nginx proxy timeout which seems to improve the subscription test which was flaky

### DIFF
--- a/config/solana-test/nano-config.toml
+++ b/config/solana-test/nano-config.toml
@@ -1,6 +1,6 @@
 [config]
-warmup_time = 50
-cooldown_time = 15
+warmup_time = 100
+cooldown_time = 20
 epoch_time = 60
 max_round_train_time = 15
 round_witness_time = 5


architectures/decentralized/testing/src/docker_setup.rs
```diff
@@ -90,10 +90,14 @@ pub async fn e2e_testing_setup_subscription(
     #[cfg(not(feature = "python"))]
     let config_file_path = ConfigBuilder::new()
         .with_num_clients(init_num_clients)
+        .with_warmup_time(15)
+        .with_round_witness_time(20)
```
entropidelic, line 94: are we hardcoding this for all tests? if we want that, they should be set directly in the `nano-config`. Regardless of that, 15 of warmup time seems very little and 20 of witness time is very large

PR #567, opened 2026-02-13
Fallback RPC retries

IAvecilla:
This PR introduces retries using all the declared RPCs. All Solana transactions made in the backend are now retryable, we retry up to 3 times with each RPC before aborting the transaction entirely.

### DIFF
--- a/Dockerfile.gateway
+++ b/Dockerfile.gateway
@@ -1,0 +1,13 @@
+# Nix builder
+FROM nixos/nix:latest AS builder
+
+# Copy our source and setup our working dir.
+COPY . /tmp/build
+WORKDIR /tmp/build
+
+RUN echo extra-substituters = https://cache.garnix.io >> /etc/nix/nix.conf
+RUN echo extra-trusted-public-keys = cache.garnix.io:CTFPyKSLcx5RMJKfLo5EEPUObbA78b0YQ2DTCJXqr9g= >> /etc/nix/nix.conf
+
+RUN nix build --extra-experimental-features nix-command --extra-experimental-features flakes .#bin-psyche-inference-node-gateway-node
+
+ENTRYPOINT nix run --extra-experimental-features nix-command --extra-experimental-features flakes .#bin-psyche-inference-node-gateway-node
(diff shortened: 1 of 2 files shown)


architectures/decentralized/solana-client/src/main.rs
```diff
@@ -216,20 +216,16 @@ async fn async_main() -> Result<()> {
...
-                backup_clusters.push(Cluster::Custom(rpc, ws))
-            }
+            let backup_clusters: Vec<_> = [(rpc_2, ws_rpc_2), (rpc_3, ws_rpc_3)]
+                .into_iter()
+                .filter_map(|(rpc, ws)| {
+                    if rpc.is_empty() && ws.is_empty() {
+                        None
+                    } else {
+                        Some(Cluster::Custom(rpc, ws))
+                    }
```
entropidelic: dont really know if it is really a valid case but what would happen if one of `rpc` or `ws` is empty?

IAvecilla: Yeah, it didn’t really make sense to keep it that way. Initially I thought someone might add an RPC without a WS and still use it as a fallback, but in practice if you have an RPC URL you also get the WS URL for free. So it’s cleaner and more consistent to require both and handle them together.

architectures/decentralized/solana-common/src/backend.rs
```diff
@@ -540,41 +569,54 @@ impl SolanaBackend {
...
-        // TODO (vbrunet) - can we improve the retry mechanism here
-        let mut retries = 0;
-        loop {
-            for program_coordinator in program_coordinators {
-                let mut request = program_coordinator.request();
-                for instruction in instructions {
-                    request = request.instruction((*instruction).clone());
+    /// Tries the operation against each RPC in order, with max retries per RPC.
+    /// Only falls back to the next RPC on retryable errors (network/timeout).
+    async fn rpc_with_fallback<T, F, Fut>(&self, f: F) -> Result<T>
```
pefontana: rpc_with_fallback always logs "rpc_with_fallback" as the operation name. Could you add a name: &str parameter and forward
  it to retry_function_with_params, like send_and_retry_with used to do, so logs show which operation actually failed?

IAvecilla: Yeah, that's actually a good suggestion

PR #569, opened 2026-02-13
[INFERENCE] Push-based model loading

samherring99:
This PR adds the following changes to the inference node setup to provide the ability to determine which model is being hosted on inference nodes:

- By default, we load inference nodes in 'idle' mode - broadcasting their status as idle to the network using iroh gossip
- We now expose an API endpoint at `/admin/load-model` from the gateway node. This is used to update which model nodes are serving. (Later on we'll add count-based tracking and routing to determine which nodes are hosting which model) - TBD on how this is most effectively used
- We add smart routing to allow for inference messages to be routed to non-idle nodes only.
- As of right now this setup ONLY uses HuggingFace checkpoints, but this will be expanded in the future
- We add a new testing command with `just test-model-loading <hf_model_name>`

## Examples of message types:

```
InferenceGossipMessage::NodeAvailable {
    model_name: Some("gpt2"),
    checkpoint_id: None,
    capabilities: vec![]
}

// Node in idle mode (no model)
InferenceGossipMessage::NodeAvailable {
    model_name: None,
    checkpoint_id: None,
    capabilities: vec![]
}

// Dynamic model loading broadcast (from gateway)
InferenceGossipMessage::LoadModel {
    model_name: "meta-llama/Llama-3.2-1B-Instruct",
    model_source: ModelSource::HuggingFace("meta-llama/Llama-3.2-1B-Instruct")
}
```

## Testing

See bullet above, but `just test-model-loading meta-llama/Llama-3.2-1B-Instruct` from inside the `nix develop .#dev-python` shell should work to demonstrate the capabilities. The test loads one model with Llama-3.2-1B-Instruct, and the other is idle. Test requests get routed to the active node. Once the `/admin/load-model` endpoint is called with gpt2, both nodes will now be hosting that model, and requests will be split between them.

### DIFF
--- a/architectures/inference-only/inference-node/src/bin/gateway-node.rs
+++ b/architectures/inference-only/inference-node/src/bin/gateway-node.rs
@@ -99,21 +99,23 @@ fn default_top_p() -> Option<f64> {
     Some(1.0)
 }
 
-#[derive(serde::Deserialize, Debug, Clone, Copy, PartialEq, Eq, Default)]
-#[serde(rename_all = "lowercase")]
-enum ModelSourceType {
-    #[default]
-    HuggingFace,
-    Local,
+#[derive(serde::Deserialize, Debug, Clone)]
+#[serde(tag = "source_type", rename_all = "lowercase")]
+enum LoadModelSource {
+    #[serde(rename = "huggingface")]
+    HuggingFace {
+        source_path: Option<String>,
+    },
+    Local {
+        source_path: String,
+    },
 }
 
 #[derive(serde::Deserialize)]
 struct LoadModelRequest {
     model_name: String,
-    #[serde(default)]
-    source_type: ModelSourceType,
-    #[serde(default)]
-    source_path: Option<String>,
+    #[serde(flatten)]
+    source: LoadModelSource,
 }
 
 #[derive(serde::Serialize)]
@@ -140,8 +142,10 @@ async fn handle_inference(
 ) -> Result<Json<ChatCompletionResponse>, AppError> {
     let nodes = state.available_nodes.read().await;
 
-    // Filter nodes that have a model loaded (not idle)
-    let nodes_with_model: Vec<_> = nodes.values().filter(|n| n.model_name.is_some()).collect();
+    let nodes_with_model: Vec<(EndpointId, String)> = nodes
+        .values()
+        .filter_map(|n| Some((n.peer_id, n.model_name.clone()?)))
+        .collect();
 
     if nodes_with_model.is_empty() {
         // No nodes have models loaded yet
@@ -150,19 +154,15 @@ async fn handle_inference(
 
     // Select first available node with a model
     // TODO: Add load balancing and model-specific routing in the future
-    let node = nodes_with_model[0];
-    let target_peer_id = node.peer_id;
+    let (target_peer_id, node_model_name) = &nodes_with_model[0];
+    let target_peer_id = *target_peer_id;
 
-    let model_name = req
-        .model
-        .clone()
-        .or_else(|| node.model_name.clone())
-        .unwrap_or_else(|| "unknown".to_string());
+    let model_name = req.model.clone().unwrap_or_else(|| node_model_name.clone());
 
     info!(
         "Routing request to node: {} (model: {})",
         target_peer_id.fmt_short(),
-        node.model_name.as_deref().unwrap_or("unknown")
+        node_model_name
     );
     drop(nodes);
 
@@ -239,20 +239,15 @@ async fn handle_load_model(
 
     info!(
         "Admin API: Received LoadModel request for model: {} (source: {:?})",
-        req.model_name, req.source_type
+        req.model_name, req.source
     );
 
-    let model_source = match req.source_type {
-        ModelSourceType::HuggingFace => {
-            let path = req.source_path.unwrap_or_else(|| req.model_name.clone());
+    let model_source = match req.source {
+        LoadModelSource::HuggingFace { source_path } => {
+            let path = source_path.unwrap_or_else(|| req.model_name.clone());
             ModelSource::HuggingFace(path)
         }
-        ModelSourceType::Local => {
-            let path = req.source_path.ok_or_else(|| {
-                AppError::BadRequest("source_path is required for local models".to_string())
-            })?;
-            ModelSource::Local(path)
-        }
+        LoadModelSource::Local { source_path } => ModelSource::Local(source_path),
     };
 
     let load_msg = InferenceGossipMessage::LoadModel {
@@ -280,7 +275,6 @@ enum AppError {
     NoNodesAvailable,
     Timeout,
     InternalError,
-    BadRequest(String),
 }
 
 impl IntoResponse for AppError {
@@ -298,7 +292,6 @@ impl IntoResponse for AppError {
                 StatusCode::INTERNAL_SERVER_ERROR,
                 "Internal server error".to_string(),
             ),
-            AppError::BadRequest(msg) => (StatusCode::BAD_REQUEST, msg),
         };
         (status, message).into_response()
     }
@@ -524,11 +517,7 @@ async fn run_gateway() -> Result<()> {
                                     InferenceGossipMessage::NodeAvailable { model_name, checkpoint_id, capabilities } => {
                                         info!("Discovered inference node!");
                                         info!("  Peer ID: {}", peer_id.fmt_short());
-                                        if let Some(ref model) = model_name {
-                                            info!("  Model: {}", model);
-                                        } else {
-                                            info!("  Model: <idle>");
-                                        }
+                                        info!("  Model: {}", model_name.as_deref().unwrap_or("<idle>"));
                                         info!("  Checkpoint: {:?}", checkpoint_id);
                                         info!("  Capabilities: {:?}", capabilities);
 
--- a/architectures/inference-only/inference-node/src/bin/test-network.rs
+++ b/architectures/inference-only/inference-node/src/bin/test-network.rs
@@ -147,11 +147,7 @@ async fn main() -> Result<()> {
                                 peer_count += 1;
                                 info!("PEER DISCOVERED!");
                                 info!("  Peer ID: {}", peer_id.fmt_short());
-                                if let Some(ref model) = model_name {
-                                    info!("  Model: {}", model);
-                                } else {
-                                    info!("  Model: <idle>");
-                                }
+                                info!("  Model: {}", model_name.as_deref().unwrap_or("<idle>"));
                                 info!("  Checkpoint: {:?}", checkpoint_id);
                                 info!("  Capabilities: {:?}", capabilities);
                                 info!("  Total peers seen: {}", peer_count);
--- a/architectures/inference-only/inference-node/src/main.rs
+++ b/architectures/inference-only/inference-node/src/main.rs
@@ -25,6 +25,13 @@ use tracing::{debug, error, info, warn};
 #[global_allocator]
 static GLOBAL: tikv_jemallocator::Jemalloc = tikv_jemallocator::Jemalloc;
 
+#[derive(Debug, Clone)]
+enum ModelLoadState {
+    Idle,
+    Loading(String),
+    Loaded(String),
+}
+
 #[derive(Parser, Debug)]
 #[command(name = "psyche-inference-node")]
 struct Cli {
@@ -109,11 +116,10 @@ async fn main() -> Result<()> {
     };
 
     info!("Starting Psyche Inference Node");
-    if let Some(ref model) = run_args.model_name {
-        info!("Model: {}", model);
-    } else {
-        info!("Model: <idle - will load on request>");
-    }
+    info!(
+        "  Model: {}",
+        run_args.model_name.as_deref().unwrap_or("<idle>")
+    );
     info!("Tensor Parallel Size: {}", run_args.tensor_parallel_size);
     info!(
         "GPU Memory Utilization: {}",
@@ -167,8 +173,11 @@ async fn main() -> Result<()> {
         Arc::new(RwLock::new(None))
     };
 
-    let current_model_name = Arc::new(RwLock::new(run_args.model_name.clone()));
-    let is_loading = Arc::new(RwLock::new(false));
+    let model_state = Arc::new(RwLock::new(if let Some(ref model) = run_args.model_name {
+        ModelLoadState::Loaded(model.clone())
+    } else {
+        ModelLoadState::Idle
+    }));
     let tensor_parallel_size = run_args.tensor_parallel_size;
     let gpu_memory_utilization = run_args.gpu_memory_utilization;
 
@@ -213,7 +222,10 @@ async fn main() -> Result<()> {
     tokio::time::sleep(Duration::from_secs(2)).await;
 
     // announce availability via gossip
-    let model_name_for_broadcast = current_model_name.read().await.clone();
+    let model_name_for_broadcast = match &*model_state.read().await {
+        ModelLoadState::Loaded(name) => Some(name.clone()),
+        _ => None,
+    };
     let availability_msg = InferenceGossipMessage::NodeAvailable {
         model_name: model_name_for_broadcast.clone(),
         checkpoint_id: None,
@@ -224,11 +236,10 @@ async fn main() -> Result<()> {
         .broadcast(&availability_msg)
         .context("Failed to broadcast availability")?;
 
-    if let Some(ref model) = model_name_for_broadcast {
-        info!("Broadcasted availability to network (model: {})", model);
-    } else {
-        info!("Broadcasted idle status to network (no model loaded)");
-    }
+    info!(
+        "Broadcasted availability to network (model: {})",
+        model_name_for_broadcast.as_deref().unwrap_or("<idle>")
+    );
     info!("Inference node ready! Listening for requests...");
 
     // heartbeat for re-announcing availability
@@ -248,7 +259,10 @@ async fn main() -> Result<()> {
             }
 
             _ = heartbeat_interval.tick() => {
-                let model_name_for_broadcast = current_model_name.read().await.clone();
+                let model_name_for_broadcast = match &*model_state.read().await {
+                    ModelLoadState::Loaded(name) => Some(name.clone()),
+                    _ => None,
+                };
                 let availability_msg = InferenceGossipMessage::NodeAvailable {
                     model_name: model_name_for_broadcast.clone(),
                     checkpoint_id: None,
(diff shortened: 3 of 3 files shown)


architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -196,6 +231,70 @@ async fn handle_inference(
...
+            ModelSource::HuggingFace(path)
+        }
+        "local" => {
+            if let Some(path) = req.source_path {
+                ModelSource::Local(path)
+            } else {
+                return Json(LoadModelResponse {
+                    success: false,
+                    message: "source_path is required for local models".to_string(),
+                });
```
IAvecilla: Maybe instead of using the `success` field we can change this function to return a `Result` and return `Err(..)` in these cases. That way we return the correct HTTP status codes for errors, I believe Axum handles that automatically. Right now we always return `200` even when there are failures.

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,84 @@ async fn main() -> Result<()> {
...
+                                    ModelSource::HuggingFace(ref name) => name.clone(),
+                                    ModelSource::Local(ref path) => path.clone(),
+                                };
+
+                                let current = current_model_name.read().await;
+                                if current.as_ref() == Some(&requested_model) {
+                                    info!("Model {} already loaded, skipping", requested_model);
+                                    drop(current);
+                                } else {
+                                    drop(current);
```
IAvecilla: Instead of dropping manually here in both cases you can use a scoped block:
```
  let model_already_loaded = {
      let current = current_model_name.read().await;
      current.as_ref() == Some(&requested_model)
  }; // Dropped here
  if model_already_loaded {
      info!("Model {} already loaded, skipping", requested_model);
      ...
  } else {
     ...
  }
```

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,84 @@ async fn main() -> Result<()> {
...
+                                    {
+                                        let mut node_guard = inference_node_shared.write().await;
+                                        if let Some(mut old_node) = node_guard.take() {
+                                            info!("Shutting down existing model");
+                                            if let Err(e) = old_node.shutdown() {
+                                                error!("Error shutting down old model: {:#}", e);
+                                            }
+                                            // Give vLLM time to release GPU memory before loading new model
+                                            // This prevents OOM when switching between large models
+                                            drop(node_guard);
```
IAvecilla: You can also avoid the manual drop here:
```
let old_node = inference_node_shared.write().await.take();
if let Some(mut old_node) = old_node {
      ...
} // Dropped here
```

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,84 @@ async fn main() -> Result<()> {
...
+                                            info!("Waiting 5s for GPU memory to be released...");
+                                            tokio::time::sleep(Duration::from_secs(5)).await;
+                                        }
+                                    }
+
+                                    match async {
```
IAvecilla: Is this async clause needed? the main function is already async and both `InferenceNode::new` and `new_node.initialize` doesn't need an async runtime so maybe we should check its removal

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,84 @@ async fn main() -> Result<()> {
...
+                                            Some(gpu_memory_utilization),
+                                        );
+
+                                        new_node.initialize(
+                                            Some(tensor_parallel_size),
+                                            Some(gpu_memory_utilization),
+                                        )?;
+
+                                        *inference_node_shared.write().await = Some(new_node);
+                                        *current_model_name.write().await = Some(requested_model.clone());
```
IAvecilla: Maybe this isn’t super important, but we could end up in a situation where `inference_node_shared` is updated correctly while `current_model_name` fails, leaving us in an inconsistent state. We could wrap both of them in a single `Arc<RwLock<(_, _)>>` so we either update both together or none at all, and avoid those inconsistencies.

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,84 @@ async fn main() -> Result<()> {
...
+                                        let mut new_node = InferenceNode::new(
+                                            model_path.clone(),
+                                            Some(tensor_parallel_size),
+                                            Some(gpu_memory_utilization),
+                                        );
+
+                                        new_node.initialize(
+                                            Some(tensor_parallel_size),
+                                            Some(gpu_memory_utilization),
+                                        )?;
```
IAvecilla: Maybe I need a bit more context to know if this is an accurate suggestion, but are these operations too heavy or likely to take some time to complete? My concern is that running this, plus the `tokio::time::sleep(Duration::from_secs(5)).await`, will block the `tokio::select` you’re using here. That would block all the heartbeats during that time.

If you don’t need the output immediately, maybe you could move all of this into a separate `tokio::task::spawn` and let it run in the background. That said, feel free to decide what makes the most sense here.

shared/inference/src/protocol.rs
```diff
@@ -201,11 +212,59 @@ mod tests {
...
+
+    #[test]
+    fn test_model_source_variants() {
+        let hf = ModelSource::HuggingFace("meta-llama/Llama-2-7b".to_string()); // make hermes
+        let bytes = postcard::to_stdvec(&hf).unwrap();
+        let parsed: ModelSource = postcard::from_bytes(&bytes).unwrap();
+        assert_eq!(parsed, hf);
+
+        // local
+        // let local = ModelSource::Local("/path/to/model".to_string());
```
IAvecilla: Can we remove this comments?

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -457,6 +574,7 @@ async fn run_gateway() -> Result<()> {
 
     let app = Router::new()
         .route("/v1/chat/completions", post(handle_inference))
+        .route("/admin/load-model", post(handle_load_model))
```
IAvecilla, line 560: Should we add some authorization for this endpoint? Don't exactly know if it's a problem to expose this

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -121,14 +140,30 @@ async fn handle_inference(
...
+        .clone()
+        .or_else(|| node.model_name.clone())
+        .unwrap_or_else(|| "unknown".to_string());
+
     info!(
         "Routing request to node: {} (model: {})",
-        node.peer_id.fmt_short(),
-        node.model_name
+        target_peer_id.fmt_short(),
+        node.model_name.as_ref().unwrap_or(&"unknown".to_string())
```
IAvecilla: We can avoid the String allocation in the log
```
node.model_name.as_deref().unwrap_or("unknown")
```

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -97,6 +98,24 @@ fn default_temperature() -> Option<f64> {
 fn default_top_p() -> Option<f64> {
     Some(1.0)
 }
+#[derive(serde::Deserialize)]
+struct LoadModelRequest {
+    model_name: String,
+    #[serde(default = "default_model_source_type")]
+    source_type: String, // "huggingface" or "local"
```
IAvecilla: Can we do an enum for this instead of using a String?

architectures/inference-only/inference-node/src/bin/gateway-node.rs
samherring99: Great callout, I'll update with this!

architectures/inference-only/inference-node/src/main.rs
samherring99: Will check removal and update here.

architectures/inference-only/inference-node/src/main.rs
samherring99: I like this suggestion too, will look into wrapping them in `Arc<RwLock<(_, _)>>` to avoid inconsistent behaviors

architectures/inference-only/inference-node/src/main.rs
samherring99: Let me look into if the select / heartbeat gets blocked here. Worst case running in background seems like a fine alternative.

shared/inference/src/protocol.rs
samherring99: Yeah, I left these in to spark any conversation about how to pull checkpoints thru GCP and iroh blobs (also so I wouldn't forget myself 😄 ) but I'll remove

architectures/inference-only/inference-node/src/bin/gateway-node.rs
samherring99: Yes we will need auth for this 🙃  - this will be a later issue but definitely one we can't forget.

architectures/inference-only/inference-node/src/bin/gateway-node.rs
samherring99: Sure, I'd set as string since it was only the two types but when we get to GCP and iroh blob type sources it makes much more sense to use an enum, so will do it now.

architectures/inference-only/inference-node/src/main.rs
samherring99: Ok actually instead I just switched the order here. it makes more sense that the node would remain in idle state if the update failed vs. advertising an incorrect model name.

`current_model_name` is only used for broadcasting the model name in gossip messages, while `inference_node_shared` is the actual inference engine, so this is transactional enough to not cause any major issues.

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,87 @@ async fn main() -> Result<()> {
...
+
+                                    // Spawn background task to avoid blocking the event loop
+                                    // Model loading can take 10-60+ seconds, so we don't want to block heartbeats
+                                    let inference_node_shared_clone = inference_node_shared.clone();
+                                    let current_model_name_clone = current_model_name.clone();
+                                    let requested_model_clone = requested_model.clone();
+
+                                    tokio::spawn(async move {
+                                        // Shutdown old model if exists
+                                        let old_node = inference_node_shared_clone.write().await.take();
```
IAvecilla, line 326: Should we also do:
```
*current_model_name_clone.write().await = None
```
here?

At this point that variable contains the old model name right? Not sure if that's an issue tho so feel free to decide what's best

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,87 @@ async fn main() -> Result<()> {
...
+                                        })();
+
+                                        match load_result {
+                                            Ok(new_node) => {
+                                                // update model name first, then node, to maintain consistency
+                                                *current_model_name_clone.write().await = Some(requested_model_clone.clone());
+                                                *inference_node_shared_clone.write().await = Some(new_node);
+
+                                                info!("Successfully loaded model: {}", requested_model_clone);
+                                                // Note: NodeAvailable will be broadcast on next heartbeat (every 30s)
```
IAvecilla, line 360: Is there a reason why we wait to the next heartbeat to announce this? Can we announce at this exact time since we are sure that the node has the model at this point

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -247,16 +266,87 @@ async fn main() -> Result<()> {
...
                         match msg {
                             InferenceGossipMessage::NodeAvailable { model_name, checkpoint_id, capabilities } => {
-                                info!("Peer {} is available: model={}, checkpoint={:?}, caps={:?}",
+                                info!("Peer {} is available: model={:?}, checkpoint={:?}, caps={:?}",
                                       peer_id.fmt_short(), model_name, checkpoint_id, capabilities);
                             }
                             InferenceGossipMessage::NodeUnavailable => {
                                 info!("Peer {} is no longer available", peer_id.fmt_short());
                             }
+                            InferenceGossipMessage::LoadModel { model_name: requested_model, model_source } => {
```
IAvecilla, line 293: Maybe this doesn’t work this way, but could it happen that we receive multiple `LoadModel` messages very close to each other?

I’m thinking about a scenario where the node starts loading the model in a `tokio::task`, and while that’s still in progress we receive another `LoadModel` message. That second one would then start loading again, recreate the inference node, and try to initialize it while the first task hasn’t finished yet. Could that potentially overfill the GPU or cause other issues?

Not sure if this is a realistic case in practice, but maybe we should add some kind of guard to indicate that a model load is already in progress and prevent it from happening twice concurrently.

architectures/inference-only/inference-node/src/main.rs
samherring99: This is a great callout I missed and actually does contain the old model name during the transition in my testing. Updating with this.

architectures/inference-only/inference-node/src/main.rs
samherring99: So we couldn't clone the network object to use it in the spawned task, and don't have access to the network object to broadcast at this exact time. For now, I'm fine leaving it for 30sec and updating this in a later PR, if that makes sense to you.

architectures/inference-only/inference-node/src/main.rs
samherring99: Good call added this for future cases. the gateway node actually handles this but its smart to have in any case.

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -121,14 +139,30 @@ async fn handle_inference(
     Json(req): Json<ChatCompletionRequest>,
 ) -> Result<Json<ChatCompletionResponse>, AppError> {
     let nodes = state.available_nodes.read().await;
-    let node = nodes.values().next().ok_or(AppError::NoNodesAvailable)?;
 
+    // Filter nodes that have a model loaded (not idle)
+    let nodes_with_model: Vec<_> = nodes.values().filter(|n| n.model_name.is_some()).collect();
```
ethernet8023: the "unknown" case for model name is not reachable here, since you verify that `model_name.is_some()`.

you could replace the `unwrap_or_else(` on 160 with a straight unwrap(), same with on 165 - 
or, better, do a 

```rs
    let nodes_with_model: Vec<_> = nodes
        .values()
        .filter_map(|n| Some((n.peer_id, n.model_name.clone()?)))
        .collect();
```

or something that lets you encode the fact that you have a model set in the type system instead of making it something you need to manually check when modifying this code

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -196,22 +230,75 @@ async fn handle_inference(
...
+    );
+
+    let model_source = match req.source_type {
+        ModelSourceType::HuggingFace => {
+            let path = req.source_path.unwrap_or_else(|| req.model_name.clone());
+            ModelSource::HuggingFace(path)
+        }
+        ModelSourceType::Local => {
+            let path = req.source_path.ok_or_else(|| {
+                AppError::BadRequest("source_path is required for local models".to_string())
```
ethernet8023: i'd consider encoding this in the enum, having

```rs
enum ModelSource {
  HuggingFace {
    source_path: Option<String>
  },
  Local {
    source_path: String
  }
```

so that the check is handled by the compiler instead of us having to remember to impl conditions

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -420,7 +518,11 @@ async fn run_gateway() -> Result<()> {
                                     InferenceGossipMessage::NodeAvailable { model_name, checkpoint_id, capabilities } => {
                                         info!("Discovered inference node!");
                                         info!("  Peer ID: {}", peer_id.fmt_short());
-                                        info!("  Model: {}", model_name);
+                                        if let Some(ref model) = model_name {
```
ethernet8023: can use fmt string like
`info!("  Model: {}", model_name.unwrap_or("<idle>");`

architectures/inference-only/inference-node/src/bin/test-network.rs
```diff
@@ -147,14 +147,22 @@ async fn main() -> Result<()> {
                                 peer_count += 1;
                                 info!("PEER DISCOVERED!");
                                 info!("  Peer ID: {}", peer_id.fmt_short());
-                                info!("  Model: {}", model_name);
+                                if let Some(ref model) = model_name {
```
ethernet8023: same format!( style thing as above

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -106,10 +109,12 @@ async fn main() -> Result<()> {
         None => cli.run_args,
     };
 
-    let model_name = run_args.model_name.context("--model-name is required")?;
-
     info!("Starting Psyche Inference Node");
-    info!("Model: {}", model_name);
+    if let Some(ref model) = run_args.model_name {
```
ethernet8023: same format! thing as above  :)

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -200,17 +214,22 @@ async fn main() -> Result<()> {
...
+        checkpoint_id: None,
         capabilities: capabilities.clone(),
     };
 
     network
         .broadcast(&availability_msg)
         .context("Failed to broadcast availability")?;
 
-    info!("Broadcasted availability to network");
+    if let Some(ref model) = model_name_for_broadcast {
```
ethernet8023: same format! thing

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -250,16 +271,99 @@ async fn main() -> Result<()> {
...
                                       peer_id.fmt_short(), model_name, checkpoint_id, capabilities);
                             }
                             InferenceGossipMessage::NodeUnavailable => {
                                 info!("Peer {} is no longer available", peer_id.fmt_short());
                             }
+                            InferenceGossipMessage::LoadModel { model_name: requested_model, model_source } => {
+                                info!("Received LoadModel request from {}: model={}, source={:?}",
+                                      peer_id.fmt_short(), requested_model, model_source);
+
+                                let model_path = match model_source {
```
ethernet8023: silly nit: consider
```rs
let model_path = match model_source.clone() {
  ModelSource::HuggingFace(name) | ModelSource::Local(name) => name
};
```

architectures/inference-only/inference-node/src/main.rs
```diff
@@ -250,16 +271,99 @@ async fn main() -> Result<()> {
...
+                                    ModelSource::HuggingFace(ref name) => name.clone(),
+                                    ModelSource::Local(ref path) => path.clone(),
+                                };
+
+                                let model_already_loaded = {
```
ethernet8023: what does it mean to have `is_loading` and `model_already_loaded` both true?
if this is not possible, consider writing an enum e.g.
```
enum ModelLoadState {
  Idle,
  Loading((name, cancelable load task maybe?)), // or maybe just Loading bare for now
  Loaded((name, inference_node_shared_clone))
}
```

instead of two booleans - make invalid states unrepresentable in the type system

samherring99: Good callout, this shouldn't be possible, will rewrite to be an enum

architectures/inference-only/inference-node/src/bin/gateway-node.rs
samherring99: Good callout, made this change

architectures/inference-only/inference-node/src/bin/gateway-node.rs
```diff
@@ -121,14 +141,28 @@ async fn handle_inference(
     Json(req): Json<ChatCompletionRequest>,
 ) -> Result<Json<ChatCompletionResponse>, AppError> {
     let nodes = state.available_nodes.read().await;
-    let node = nodes.values().next().ok_or(AppError::NoNodesAvailable)?;
 
-    let target_peer_id = node.peer_id;
-    let model_name = req.model.clone().unwrap_or_else(|| node.model_name.clone());
+    let nodes_with_model: Vec<(EndpointId, String)> = nodes
+        .values()
+        .filter_map(|n| Some((n.peer_id, n.model_name.clone()?)))
```
ethernet8023, line 147: 🧠 neat syntax

PR #570, opened 2026-02-13
Disable CI action for run-manager which was commented out

dsocolobsky:
There is a CI action which always fails in every push because every step is commented out.

Remove it for now and we'll create an issue once we're ready to have it again running.

image


.github/workflows/run-manager-integration-test-run.yml
```diff
@@ -12,32 +9,31 @@ jobs:
...
-          # - test_checkpoint
-          # - test_join_authorization_delegate
-          # - test_json_dump_user
-    permissions:
-      contents: read
-      packages: write
-    uses: ./.github/workflows/run-manager-integration-test-base.yml
-    with:
-      test-name: ${{ matrix.test-name }}
+  # Run-manager integration tests are temporarily disabled due to CI flakes.
```
IAvecilla: I would just delete it, I'll create an issue as a reminder for us to properly add them

dsocolobsky: ok then

PR #592, opened 2026-02-26
Configuration required to get the Gateway running on Railway

benbarclay:
In order to get the Gateway service running on Railway, the following two modifications were required:

1. Create a Dockerfile, as the Nix build wasn't a supported build option
2. Change the default network interface binding to 0.0.0.0 rather than the local loopback

With these two changes in place, we have the Gateway successfully running on Railway and handing a small amount of traffic. With this change, we can set up a full automated deployment pipeline.

The only real downside is that the build takes about ~30 minutes in Railway. Eventually we may want to build the image in GitHub Actions and push just the artifact to Railway, but that refinement can come later.

### DIFF
--- a/architectures/inference-only/inference-node/src/bin/gateway-node.rs
+++ b/architectures/inference-only/inference-node/src/bin/gateway-node.rs
@@ -34,7 +34,7 @@ use tracing::{debug, error, info};
 
 #[derive(Parser, Debug)]
 struct Args {
-    #[arg(long, default_value = "127.0.0.1:8000")]
+    #[arg(long, default_value = "0.0.0.0:8000")]
     listen_addr: String,
 
     /// what discovery to use - public n0 or local


Dockerfile.gateway
```diff
@@ -0,0 +1,13 @@
...
+# Copy our source and setup our working dir.
+COPY . /tmp/build
+WORKDIR /tmp/build
+
+RUN echo extra-substituters = https://cache.garnix.io >> /etc/nix/nix.conf
+RUN echo extra-trusted-public-keys = cache.garnix.io:CTFPyKSLcx5RMJKfLo5EEPUObbA78b0YQ2DTCJXqr9g= >> /etc/nix/nix.conf
+
+RUN nix build --extra-experimental-features nix-command --extra-experimental-features flakes .#bin-psyche-inference-node-gateway-node
+
+ENTRYPOINT nix run --extra-experimental-features nix-command --extra-experimental-features flakes .#bin-psyche-inference-node-gateway-node
```
ethernet8023, line 13: instead of keeping the entire build closure here and going thru `nix run`, what if we did
```dockerfile
# copy out the entire closure for this binary
RUN nix copy extra-experimental-features "nix-command flakes" --to /tmp/nix-store-staging .#bin-psyche-inference-node-gateway-node
RUN realpath result/bin/* > /tmp/entrypoint

# use a fresh minimal image to run the binary
FROM debian:slim
COPY --from=builder /tmp/nix-store-staging/nix/store /nix/store
COPY --from=builder /tmp/entrypoint /entrypoint
ENTRYPOINT ["sh", "-c", "exec $(cat /entrypoint)"]
```

to make nice small images w/ faster startup?
if it doesn't matter, it doesn't matter, though.

arilotter (approved):

Dockerfile.gateway
benbarclay: As we chatted about on Discord, I'll merge this as a first pass then look into using Nix to build the container directly

PR #635, opened 2026-03-13
Use `bf16` type on model sharing

IAvecilla:


ethernet8023 (approved):
approved with an Ok from Jeff - we should see about making sure checkpoints save and load from this format as well, and maybe throw an error if we try and load a checkpoint at any step other than 0 that's in any format other than fp16. thoughts?

ethernet8023 (approved):
i can't follow the barrer logic but it looks the same as the other methods, so lgtm.

PR #639, opened 2026-03-20
Run down signed urls 2

pefontana:


dsocolobsky:
LGTM I've compared it with my solution and it's pretty much the same.

